PR #7, opened 2015-08-09
Fixed a typo in config

clearjs:


davezuko:
Haha oh man, you got this PR in just before I was about to fix this. Good catch, I'll merge yours in so you get the credit. Thanks!

PR #9, opened 2015-08-09
This should allow Travis to build iojs.

clearjs:
"iojs" is an alias for the latest stable version, which is currently 3.0, and with that version the project wouldn't run locally either.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: node_js
 node_js:
   - "0.12"
+  - "iojs-v2"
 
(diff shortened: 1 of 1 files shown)


davezuko:
And here I was thinking I was on latest iojs. Do you have any idea why it might be breaking with 3.0? I'll have to go investigate that today, as well as document node version requirements in the project.

Thanks.

clearjs:
Probably not all deps have been upgraded to support it yet.

PR #11, opened 2015-08-09
fix(regex): typos

clearjs:
`?` means optional, I think it was a copy-paste typo


davezuko:
Yeah you're right, thanks!

PR #12, opened 2015-08-09
You're actually using 1.0 RC already

gaearon:


davezuko:
Oh thanks for catching that. Awesome work with Redux by the way!

PR #32, opened 2015-08-25
Fix webpack client config

nuragic:
Hi!

The require call was failing due to an extra space so, I just added a `.trim()`... :smile: 

This is my npm script that was raising the error (I'm on Windows right now):

`"deploy": "SET NODE_ENV=production && npm run test && npm run compile"`

Thanks!

### DIFF
--- a/build/webpack/client/index.js
+++ b/build/webpack/client/index.js
@@ -61,5 +61,5 @@ config.module.loaders.push(
 );
 
 module.exports = function makeClientConfig (type) {
-  return require('./_' + (type || projectConfig.NODE_ENV))(config);
+  return require('./_' + (type || projectConfig.NODE_ENV).trim())(config);
 };


davezuko:
Oh awesome, thanks so much! I never even thought about that.

Merged.

nuragic:
Thanks!

nuragic:
@davezuko Man, I just realized that it wasn't such a good fix... the `.trim()` call should be placed in this line:

https://github.com/davezuko/react-redux-starter-kit/blob/master/config/index.js#L1

Otherwise, (on windows AFAIK) you get this output:
```js
process.env.NODE_ENV // "production " 
process.env.NODE_ENV === 'development' // false
process.env.NODE_ENV === 'production' // false
``` 

I'll send another PR to fix that, it's ok?

davezuko:
Yeah sure. Is this just a Windows issue then? Because I've yet to have trouble on OSX/Ubuntu.

nuragic:
Ok I'll send it! Yes, I suppose it's just a windows issue... -.-

PR #38, opened 2015-08-31
fix(config): trim spaces from NODE_ENV to avoid problems on windows

nuragic:
See https://github.com/davezuko/react-redux-starter-kit/pull/32.


davezuko:
Thanks!

PR #40, opened 2015-09-01
Upgrade/dependencies

davezuko:


iamstarkov:
> "\<Provider\> does not support changing `store` on the fly. It is most likely that you see this error because you updated to Redux 2.x and React Redux 2.x which no longer hot reload reducers automatically. See https://github.com/rackt/react-redux/releases/tag/v2.0.0 for the migration instructions."

I got this error from the start, so probably source code for reducers also should be updated

PR #41, opened 2015-09-01
Npm start

iamstarkov:
its rather convenient to install deps and then just start the app: `npm i && npm start`. What do you think?


davezuko:
Agreed, that makes a lot of sense... and just learned something new today (`npm i`).

Merged, thanks!

PR #78, opened 2015-09-13
Add babelRc and devDeps

bulby97:


davezuko:
Thanks! I'm about to run out for a bit but I'll check it out when I get home.

davezuko:
Awesome, merged. I actually prefer the regular redbox styles since they're more noticeable, but this works great until that environment issue gets fixed. Thanks!

bulby97:
Yeah me too, I like to feel assaulted whith that big bold red when I am making a mistake haha.

PR #95, opened 2015-09-20
Use react-transform-hmr

daviferreira:
The package `react-transform-webpack-hmr` was renamed to `react-transform-hmr`. This PR simply update the dependencies/scripts.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,16 +1,18 @@
 {
   "stage"   : 0,
-  "plugins" : [
-    "react-transform"
-  ],
-  "extra": {
-    "react-transform": [{
-      "target" : "react-transform-webpack-hmr",
-      "imports" : ["react"],
-      "locals" : ["module"]
-    }, {
-      "target" : "react-transform-catch-errors",
-      "imports" : ["react", "redbox-react"]
-    }]
+  "env": {
+    "development": {
+      "plugins" : ["react-transform"],
+      "extra": {
+        "react-transform": [{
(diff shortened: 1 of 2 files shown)


davezuko:
I haven't had a chance to try out the new package locally yet; any idea about why this is build is failing? Your configuration looks correct at a glance. From the Travis log:

```
 Error: locals[0] does not appear to be a `module` object with Hot Module replacement API enabled. You should disable react-transform-hmr in production by using `env` section in Babel configuration. See the example in README: https://github.com/gaearon/react-transform-hmr
```

Thanks for this though, I'll gladly pull it in once the build is fixed.

davezuko:
Ah, it's actually because you need to explicitly define it in development:

```
  "env": {
    // only enable it when process.env.NODE_ENV is 'development' or undefined
    "development": {
      "plugins": ["react-transform"],
      "extra": {
        // must be defined and be an array
        "react-transform": [{
          "target": "react-transform-hmr",
          // if you use React Native, pass "react-native" instead:
          "imports": ["react"],
          // this is important for Webpack HMR:
          "locals": ["module"]
        }]
```

daviferreira:
Sorry I didn't check the build, my bad! Should be good now.

Thanks for this repo, checked a couple of boilerplates/starters out there, and fell in love with yours :)

davezuko:
No problem, thanks for your help :).

And that's awesome, it's good to hear some people are finding it helpful. Thanks again!

PR #104, opened 2015-09-24
Example/todo application

agdolla:
updated to current master
new tests
simplified layout (components not nested into own directory)


davezuko:
Thank you so much, I can't begin to say how much I appreciate this!

Merged.

PR #107, opened 2015-09-24
[production-clean][clean client folder pre production compile]

Casedy:


davezuko:
Hey thanks. Would you mind updating the `dist` folder reference to use the path helper provided in the config? That way in case somebody changes the name of their `dist` folder in the config everything just automagically works.

Something like this maybe?

```
// ...
new Clean(['/client'], paths.dist())
```

This could honestly probably just go in the default plugins rather than production only, since there are valid reasons to compile the app to disk outside of production. Let me know your thoughts on this. Thanks again for your help!

Casedy:
`This could honestly probably just go in the default plugins rather than production only`

Pretty good idea. I have moved this code to `!__DEV__` env. Because `npm run dev` env doesn't build new files in `client` folder.

davezuko:
Thanks for making that change, merged!

PR #120, opened 2015-10-06
Add Getting Started section to Readme

bpugh:
This is probably obvious for most React devs but might not be so straightforward
to someone not used to a Node build process. Thoughts?


davezuko:
Great idea, and looks good to me. Could you add a reference to that section in the table of contents? Happy to merge it in if you include that. Thanks!

bpugh:
Updated! Thanks!

davezuko:
Thanks, merged!

PR #121, opened 2015-10-06
fix(webpack): do not generate source maps in production builds

### LINKED ISSUE #119: Production builds should not contain source maps
Currently source maps are generated for development builds as well as for production builds. I think it would be better not to generate source maps in a production build by default.

Setting the devtool option in build/webpack/development.js instead of build/webpack/_base.js should fix this.

What do you think?

kiejo:
This closes #119


davezuko:
Haha easy enough, thanks for making that change so quickly. Merged!

PR #123, opened 2015-10-09
Ability to add json-loader

brancusi:
Updates react-transform HMR check to not match json-loader.

While trying to use the color-loader package, I ran into an issue with the react-transform HMR map function, matching when including the follow loader block:
```js
{
  test: /\.json$/,
  loader: 'json-loader'
}
```

Since I'm assuming it was only meant to match the babel loader.

Lastly, is there a reason you match on the test and not the loader prop?

Example: 

```js
if (/babel/.test(loader.loader)) { ... }
```
Seems to work as well?

Thanks

### DIFF
--- a/build/webpack/development_hot.js
+++ b/build/webpack/development_hot.js
@@ -17,7 +17,7 @@ webpackConfig.plugins.push(
 // configuration will break other tasks such as test:unit because Webpack
 // HMR is not enabled there, and these transforms require it.
 webpackConfig.module.loaders = webpackConfig.module.loaders.map(loader => {
-  if (/js/.test(loader.test)) {
+  if (/js(?!on)/.test(loader.test)) {
     loader.query.env.development.extra['react-transform'].transforms.push({
       transform : 'react-transform-hmr',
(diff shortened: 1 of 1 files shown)


davezuko:
The reason for the check on test and not loader was because (way back when) I used to have multiple loaders for js files, so it made more sense to check a singular test rather than iterate through arrays. That's not so much the case anymore, but oh well.

This looks like a much-needed fix, thank you. If you have any swaying arguments for moving the check from test to loader then I'd be happy to consider an additional PR, but I don't think it's causing too many problems as it stands.

brancusi:
Thanks @davezuko, glad it's a fix. I really don't understand the need enough to say if the loader.loader makes more sense. If it only ever needs to match the babel loader, then it might be safer, in case someone needs to add some other loader that contains a **`js`** in the test regex. Nothing comes to mind at the moment but I imagine there are cases.

I can make a PR if you think it's a better approach. I think you mention in the source that it's a temp workaround, maybe not worth touching it anymore since it may get refactored anyway.

Thanks

Issue #126, opened 2015-10-14
Local bootstrap?


OKNoah:
I see that the Bootstrap CDN is used. How can I install it right in the project?

I've installed Bootstrap and Bootstrap-sass, but I get errors like these,

    ERROR in ./~/bootstrap-sass/assets/fonts/bootstrap/glyphicons-halflings-regular.woff
    Module parse failed: /Users/Username/Documents/dev/functin-frontend/node_modules/bootstrap-sass/assets/fonts/bootstrap/glyphicons-halflings-regular.woff Line 1: Unexpected token ILLEGAL
    You may need an appropriate loader to handle this file type.
    (Source code omitted for this binary file)
     @ ./~/css-loader!./~/autoprefixer-loader?browsers=last 2 version!./~/sass-loader?includePaths[]=/Users/Username/Documents/dev/functin-frontend/src/styles!./~/bootstrap-sass/assets/stylesheets/_bootstrap.scss 6:4446-4509

nodkz:
You should add loaders for fonts files.

For example, this is mine part of config with extended params: 
`
config['module']['loaders'] = [
    ....
    { test: /\.(jpg|png)$/,       loader: 'url-loader?name=[path][name].[ext]&limit=100000' },
    { test: /\.woff(\?.*)?$/,     loader: "url-loader?prefix=fonts/&name=[path][name].[ext]&limit=10000&mimetype=application/font-woff" },
    { test: /\.woff2(\?.*)?$/,    loader: "url-loader?prefix=fonts/&name=[path][name].[ext]&limit=10000&mimetype=application/font-woff2" },
    { test: /\.ttf(\?.*)?$/,      loader: "url-loader?prefix=fonts/&name=[path][name].[ext]&limit=10000&mimetype=application/octet-stream" },
    { test: /\.eot(\?.*)?$/,      loader: "file-loader?prefix=fonts/&name=[path][name].[ext]" },
    { test: /\.svg(\?.*)?$/,      loader: "url-loader?prefix=fonts/&name=[path][name].[ext]&limit=10000&mimetype=image/svg+xml" }
];
`

davezuko:
@OKNoah I just merged in @nodkz 's PR (thank you!) so give it a try again and see if you're still having problems.

mikedevita:
I am having "funkiness" with this, probably due to user error.. but either way.

```
styles/vendor/_bootstrap.css
```

and then in my `styles/core.scss` i have
```
[...]

PR #127, opened 2015-10-15
Add support for fonts files in webpack

nodkz:


nodkz:
https://github.com/davezuko/react-redux-starter-kit/issues/126

nodkz:
According to issue https://github.com/davezuko/react-redux-starter-kit/issues/126

davezuko:
Great, thank you. I don't want to make bootstrap a local dependency, but this will be great for anyone who does (as per the issue you referenced). Thanks!

PR #130, opened 2015-10-17
Host/port config cleanup

apaatsio:
Just some small cleanup

- Remove unused config (server_port)
- Consistent naming and support PORT env var


davezuko:
Thanks! Some of these were leftovers from back when the starter kit came with a Koa server, so it's nice to clean them up. Merged.

PR #133, opened 2015-10-21
Add Redux Router.

dougvk:
@davezuko two purposes here:

* Wanted to document how to switch over to redux router for anyone who is interested
* Wanted to make sure my suggested changes aren't breaking anything (please let me know!)

Thanks for your work here -- I'm switching from pure rails to rails api + react frontend and this has helped quite a bit.

### DIFF
--- a/config/index.js
+++ b/config/index.js
@@ -23,6 +23,7 @@ config.set('vendor_dependencies', [
   'react-redux',
   'react-router',
   'redux',
+  'redux-router',
   'redux-devtools',
   'redux-devtools/lib/react'
 ]);
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
     "react-redux": "^4.0.0",
     "react-router": "1.0.0-rc1",
     "redux": "^3.0.0",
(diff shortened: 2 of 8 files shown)


davezuko:
Hey thanks, I appreciate it and it looks pretty good. There's one change I'd like to make since currently nothing renders for the index route. Could we maybe do something akin to this?

```
import { Route, IndexRoute } from 'react-router';
import React           from 'react';
import CoreLayout  from 'layouts/CoreLayout';
import HomeView   from 'views/HomeView';

export default (
  <Route path='/' component={CoreLayout}>
    <IndexRoute component={HomeView} />
  </Route>
);
```

I'm not a huge fan of the random links next to the "increment" button in the HomeView. I understand it's there to demo the route functionality, but I don't really think it's all that necessary to prove that react router does its job. If you are cool with making those changes I'd love to merge this in.

dougvk:
@davezuko great! let me know how this works for you.

davezuko:
Perfect, thank you so much! Merged.

PR #137, opened 2015-10-23
Add dotenv.

dougvk:
This allows for easy modification of environment variables. I like it and use it in all my rails projects, thought you might want to include it here.


davezuko:
What about just adding the .env straight into the project? And then environment-specific tasks (such as `npm run dev`) would just make sure to specify the correct environment variables? That would get around the whole broken CI build and need to copy/paste a sample env file.

dougvk:
do you mean e.g. prefix the dev commands with `NODE_ENV=development`?

davezuko:
Yeah, we'd have to do it for all of the `npm run dev` commands as well as `npm run test:dev`, but the rest of the commands (compile, test, deploy) would all just use what's configured in .env. Does that make sense? I've personally never used a `.env` file, so I'm not sure I'm approaching this 100% correctly.

davezuko:
Well that was fast, lol. I'll merge it in as soon as the CI build completes, thanks!

dougvk:
happy to help. i'm new to all these tools and this project got me way ahead of where i would be on my own :+1:

davezuko:
Well I appreciate all the help!

PR #141, opened 2015-10-26
Add Dynamic host in webpack-dev-server

waynelkh:
It can dynamic setting host in config file


davezuko:
Thank you for adding this! Merged.

PR #149, opened 2015-10-28
Better handling of `dev:nw`, we don't need to use the dock component …

jhgg:
…if we're using the new window for the dev tools.

Additionally, configure the new window to open at a more sane size.

image

Before, the utility window would try to dock the devtools to the left. So, 70% of the page would be blank. This makes it so that the utility window is filled by the `LogMonitor`, without any of the docking functionality.


davezuko:
Perfect, thanks! I saw this but didn't have a chance to fix it, so I appreciate the help.

PR #170, opened 2015-11-09
fix(karma): use sassLoader configuration

davezuko:


wufeng87:
if I want to use less, then how to change this config? thx.

PR #185, opened 2015-11-17
replaced back tick with single quote to fix crash on osx

mbradleyis:


davezuko:
What node version you are on? You shouldn't be getting this error as long as you're at least on `io.js ^2.0.0` or any node version after that.

Also, careful because just changing it to a string literal from a template string now means those variables aren't interpolated, so it will just spit out `webpack-dev-server is now running at ${host}:${port}.`, rather than including their values.

mbradleyis:
Oh ok, I see, you're using template strings.  I updated to v4.1.2 and the back ticks don't cause any problems.  Thanks for taking a look!

davezuko:
Cool, thanks for upgrading. I know it's such a trivial thing, but I appreciate you making a PR rather than just submitting an issue!

PR #191, opened 2015-11-23
fix(fe): Use ES5 syntax instead of ES6

cidevant:
Dont use illegal ES6 syntax (template strings), which causes crash before start.

### DIFF
--- a/bin/webpack-dev-server.js
+++ b/bin/webpack-dev-server.js
@@ -1,13 +1,12 @@
 require('babel/register');
 
-const chalk     = require('chalk');
-const devServer = require('../build/webpack-dev-server');
-const config    = require('../config');
+var chalk     = require('chalk');
+var devServer = require('../build/webpack-dev-server');
+var config    = require('../config');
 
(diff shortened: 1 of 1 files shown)


davezuko:
I really didn't want to do this since it's a non-issue if you're on anything above `iojs ^2.0.0` (see my comment in https://github.com/davezuko/react-redux-starter-kit/pull/185) and version requirements are documented in the README, but honestly at this point it probably doesn't hurt anyone to just stick to ES5 everywhere that isn't preprocessed with Babel.

If you could update `~/webpack.config.js` to not use `const` to match these changes I'd be happy to merge this in.

cidevant:
valid arguments, will fix it. The reason I created this pull request, coz app didn't work after clear `npm install`. Nothing is perfect in this world ;-)

davezuko:
Ok cool, just let me know when you've made those updates and we can get it merged in.

davezuko:
Ended up fixing this with https://github.com/davezuko/react-redux-starter-kit/pull/222

PR #194, opened 2015-11-25
Use postcss-loader to plug in autoprefixer

neverfox:
Accomplishes the same thing as the specialized autoprefixer-loader, but now the kit is set up for the addition/configuration of other postcss plugins, as desired.


davezuko:
Very awesome, thank you!

PR #198, opened 2015-11-26
(refactor)app: switch to redux-simple-router

### LINKED ISSUE #196: Switch to redux-simple-router

justingreenberg:
resolves #196

includes minimal demo `/about` route with basic [stateless function component `AboutView`](https://github.com/justingreenberg/react-redux-starter-kit/blob/master/src/views/AboutView.js)

-

**screenshots**
  image
click on **Go To About View** which is a react-router `Link` [component](https://github.com/davezuko/react-redux-starter-kit/pull/198/files#diff-def9b7b3e3020460b0cb272526a5877cR35)
  image 
`redux-simple-router` keeps the path [synchronized](https://github.com/davezuko/react-redux-starter-kit/commit/fc505bf464c650b5c81c549cbb3c4be7da77bf6e#diff-bd9c9dcd314f2d7df52935b3a6a4d504R12) with store

-

i think there's a some confusion about this library's scope and relationship with `react-router`, so it might be a good idea to include a link to the `redux-simple-router` ie. [How To Use](https://github.com/jlongster/redux-simple-router#how-to-use):

>The idea of this library is to **use react-router's functionality exactly like its documentation tells you to**. You can access all of its APIs in routing components. Additionally, you can use redux like you normally would, with a single app state and "connected" components. It's even possible for a single component to be both if needed.

>This library provides a single point of synchronization: the routing.path piece of state which is the current URL. You can read it, and also change it with an action.

TLDR: use `react-router` **as you normally would**!

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ Features
   * Includes react-addons-test-utils (`^0.14.0`)
 * [React-Router](https://github.com/rackt/react-router) (`^1.0.0`)
 * [Redux](https://github.com/gaearon/redux) (`^3.0.0`)
-  * redux-router (`^1.0.0-beta3`)
+  * redux-simple-router (`^0.0.10`)
   * react-redux (`^4.0.0`)
   * redux-devtools
     * use `npm run dev:nw` to display in a separate window.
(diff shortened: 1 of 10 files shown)


davezuko:
Awesome, thank you! I'm gone on vacation so might not get a chance to look through and merge until Sunday, but we'll see.

davezuko:
Actually was easier to look at the diffs on my phone than I thought, looks perfect. Thanks again!

PR #200, opened 2015-11-26
Fixed a breakage while creating the store

Shahul3D:
When debug mode is disabled by running "mpn run dev:no-debug", createStoreWithMiddleware function fails since its not initialized.


davezuko:
Thank you for catching this!

PR #206, opened 2015-12-01
Simplifies how action creators are passed to the HomeView

davidgtonge:
I think as a starter project its a good idea to push people towards the more simple, but elegant use of the `connect` api from `react-redux`

### DIFF
--- a/src/views/HomeView.js
+++ b/src/views/HomeView.js
@@ -4,7 +4,7 @@ import { connect }            from 'react-redux';
 import counterActions         from 'actions/counter';
 import { Link }               from 'react-router';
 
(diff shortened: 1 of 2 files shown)


mtsr:
I like it.

Seems you forgot the CI is set up to fail on lint too.

davezuko:
@davidgtonge if you could fix the lint errors I'll get this merged in. Also, if you wouldn't mind removing the extraneous `routerState` selection in `mapStateToProps` I'd appreciate it; kill two birds with one stone.

davidgtonge:
@davezuko thanks have fixed the lint errors and removed the unused router state

davezuko:
Great, thanks. Merged.

Issue #204, opened 2015-12-01
1.0.0 Roadmap


davezuko:
While a `1.0.0` release isn't exactly meaningful for a starter kit, it would be nice to finally have the kit stabilize. As such, here's what I'm thinking we need in order to meet this target:

* [x] Upgrade Babel from `^5.0.0` to `^6.0.0` (https://github.com/davezuko/react-redux-starter-kit/issues/150)
  * [x] Verify hot reloading works with react-transform
  * [x] Verify redbox-react / react-transform-catch-errors works (I'd be OK with dropping this if needed)
* [x] Implement css-modules https://github.com/davezuko/react-redux-starter-kit/issues/195
* [x] Add support for static assets
  * [x] Add favicon
* [x] Update all dependencies and devDependencies
  * [x] bump redux-simple-router to `^1.0.0`
* [x] Refactor sample app to follow ducks-like structure
* [x] Cleanup documentation
  * [x] Document express server
  * [x] Improve configuration section
  * [x] Verify all mentioned packages are accurate
  * [x] Document general deployment guidelines
* [x] Add cssnano plugin to postcss configuration
* [x] Replace autoprefixer with cssnano configuration
* [x] Replace all `console.log` with debug package (https://github.com/davezuko/react-redux-starter-kit/issues/218)

@neverfox suggested the css-modules addition, which would be awesome to add, and @justingreenberg has been keeping an eye on the Babel 6 transition so overall it looks fairly covered. Thanks for everyone's contributions, suggestions, and hard work, it's incredibly appreciated.

mtsr:
I have css-modules working in combination with react-css-modules. I could make a PR as well, if desired.

davezuko:
@mtsr that would be great, but would you be able to submit the PR's separately? I really want to take a look at the react-css-modules documentation before committing to it, so it might be helpful to get css-modules in and then possibly add react-css-modules afterward.

mtsr:
Sure, easy enough. I use react-css-modules only for the improved syntax. It works without it, but mixing global and local classes becomes messy.

neverfox:
I think sticking to a pure css-modules webpack configuration plus use of `classnames` is a good start. I have some issues with the was `react-css-modules` currently does things that may get worked out, but it's still baking.

davezuko:
[...]

Issue #195, opened 2015-11-25
CSS Modules


neverfox:
Is there any interest in setting up the kit to use [CSS Modules](http://glenmaddern.com/articles/css-modules) instead of the basic SASS style sheets it currently uses? I know this is an opinionated area, what with all the various takes on styling React (e.g. the inline vs CSS debate). It's a fairly minor change to webpack configuration and I don't mind submitting a PR if the answer is 'yes', but I thought I would gauge the interest first.

davezuko:
I'm not opposed to it, but at the same time I'm also trying to keep this starter kit as slim as possible so it doesn't become a kitchen sink like a lot of the other boilerplates available. That's not exactly applicable to CSS modules since they replace the Sass stylesheets, but it does (imo) add another barrier to entry for new users (myself included).

I'd be curious to see what other people have to say about this.

loklaan:
I've created a branch, with css-modules, to play around in. I'm likely to soon use it in a product, so it will be tested. :)

https://github.com/loklaan/react-redux-starter-kit/tree/css-modules

neverfox:
One of the nice things about it is that you can set it up right alongside the current SASS configuration and even use it while you're also using an inline approach like Radium or react-look. That's part of the beauty of Webpack loaders.

davezuko:
@neverfox I'd be open to a PR for this if you're up for it. I can't decide which would be better to demo though, if we keep it alongside the sass configuration (which would probably be wise).

vaselok:
Guys, are you planning to implement sass/css bundling to one file out of the box?

davezuko:
@vaselok yes, it's in the works here: https://github.com/davezuko/react-redux-starter-kit/pull/230

davezuko:
This has been implemented.

vaselok:
awesome!

PR #207, opened 2015-12-01
Added css-modules to css-loader

mtsr:
Following https://github.com/davezuko/react-redux-starter-kit/issues/204 and https://github.com/davezuko/react-redux-starter-kit/issues/195

Unsure about necessity of import-loaders, but haven't tested with @import
Added a local .btn to homeview increment button, css will show a locally scoped class.

I'm not entirely sure yet about css-modules functionality. It's not quite the same as having locally scoped css, because you need to specifically add a class from the local styles to every element you want to style. Combining this with global styles (such as bootstrap) leads to some ugly lines like `className={'btn '+styles.btn}`. This last bit can be solved by using react-css-modules, which allows for `styleName='btn'`, which is automatically resolved to the local style called `btn`, but still allows for classNames for global styles.

P.S. consider this a proof of concept. If this is something that will be added, I can make a cleaner example.

### DIFF
--- a/build/webpack/_base.js
+++ b/build/webpack/_base.js
@@ -65,7 +65,7 @@ const webpackConfig = {
         test    : /\.scss$/,
         loaders : [
           'style-loader',
-          'css-loader',
+          'css-loader?modules&importLoaders=2&localIdentName=[local]___[hash:base64:5]',
           'postcss-loader',
           'sass-loader'
         ]
--- a/src/styles/core.scss
+++ b/src/styles/core.scss
@@ -1,2 +1,2 @@
 @import 'base';
-@import 'vendor/normalize';
+:global { @import 'vendor/normalize'; }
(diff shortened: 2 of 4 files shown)


mtsr:
Thanks. Will fix.

davezuko:
Cool, thanks @neverfox and @mtsr; once that gets fixed we can get this merged.

davezuko:
@mtsr I added `sourceMap` to the css-loader query string, so make sure you rebase before making the update. Thanks!

mtsr:
Sorry for the delay, I'm currently quite busy with the prototype I'm using this starter kit for. This PR will likely be pushed to next week.

davezuko:
No problem @mtsr; if I manage to get to it before then is that alright? Not even sure I'll be able to, but in case I do I don't want to take over your contribution.

mtsr:
Sure, no worries. I'll find another opportunity to contribute ;).

I was thinking the test for modules should be anything in the components/views/layouts directories using CSS-module, only stuff from styles/ loaded without modules.

One thing that has me worried is that `@import` from a CSS-module will probably use CSS-module for the @imported stuff too. On the other hand `:global {}` should fix that.

davezuko:
Moving discussion to https://github.com/davezuko/react-redux-starter-kit/pull/230

Issue #189, opened 2015-11-20
Remove extra div around dev tools


nhagen:
I was about to submit a PR to remove the 2 divs in `Root`, and saw you took the first one out. Figured I'd have the discussion before removing the second one: When you compile for production, the div inside of `Provider` is still there, but serves no semantic purpose.

Would you be willing to change `Root` to do [this] (https://github.com/rackt/redux/blob/master/examples/real-world/containers/Root.js) instead?

davezuko:
Hey, good catch. It makes sense, but I'm a bit torn as far as the implementation of it; I find the 3 files very excessive and inelegant. I believe if you just use the `__DEV__` or `__PROD__` flags to generate the inner content of the `<Provider`>, then through minification/dead code removal in production that check will go away entirely. Thinking something like this:

```
render () {
  // this doesn't have to be a variable, it could just be inline, and it would become static
  // after it's processed with uglify.
  const content = __DEV__ ?
    ( ... ) :
    ( ... );

    return (
      <Provider>
        {content}
      </Provider>
    );
}
```

Thoughts? I think this way really helps eliminate duplicating `Root` in two separate files, but it's open for discussion.

nhagen:
I agree, and attempted that but was still seeing the dead code when I
compiled with production. I'll try it again that way and see. I agree real
world's implementation is a bit drastic.

davezuko:
Oh strange, I figured it would have been removed... I'll take a look as well when I get a chance. Let me know what you find out, since this would be a nice change to have.

davezuko:
@nhagen this seems to be working for me, what are your thoughts?

<img width="605" alt="screen shot 2015-11-23 at 10 59 30 am" src="https://cloud.githubusercontent.com/assets/6439050/11341653/574c62dc-91d1-11e5-8ef8-dcd70e05158f.png">

And this is the resulting code (beautified) from a production build:

```
{
             key: "render",
             value: function() {
                 var e = l["default"].createElement(p.ReduxRouter, null, d["default"]),
                     t = e;
                 return l["default"].createElement(c.Provider, {
                     store: this.props.store
                 }, t)
             }
         }
```

nhagen:
I've been having an unrelated issue with compiling so I haven't had a chance to play with this, but could you do something like `const content = __PROD__ || this.props.debug ? router : .... ;`

davezuko:
[...]

PR #211, opened 2015-12-02
refactor(environments): eliminate extra dev code/markup in production

davezuko:


davezuko:
Pinging @neverfox, @mtsr, and @nhagen since you are all fairly active contributors, I'd just like to get your opinion on this if you have a minute. It stems from https://github.com/davezuko/react-redux-starter-kit/issues/189 and a deeper issue where a lot of DevTool-related code still makes its way into production, so this is one sure-fire way of eliminating it. It's inelegant and a bit obtuse for people new to the starter kit, so I'm wondering if any of you share my wariness.

justingreenberg:
@davezuko no prob, i will take a look this afternoon. i was playing around with babel 6 migration last night and the only issues were with the current devtools setup, so this was going to be an problem down the road either way.

i will try the same with this PR and let you know results.

another solution i was looking into was using babel transforms to inject devtools in development using https://github.com/alexkuz/react-transform-debug-inspector

will follow up later :smile:

neverfox:
Would it not accomplish the same thing to do a conditional require of `DevTools` inside of a single `Root.js`?

davezuko:
@neverfox that could work, but if we go that route then both the store and Root are configured at compile time rather than at runtime, since `configureStore` would have to know to never import the DevTools. I wish dead code removal would remove unused imports, would make this whole issue trivial.

davezuko:
I'm going to close this because I think it complicates the starter kit, and anybody concerned with an as-slim-as-possible production bundle can figure this out on their own. Might offer an explanation for how to do this in https://github.com/davezuko/react-redux-starter-kit/issues/221, but I want this starter kit to be (relatively) approachable.

nhagen:
I think that's very reasonable, though I still favor this being optimized for production out of the box. If that's how it goes, there should definitely be a section of documentation that identifies optimizations that might be useful in production environments (like this)... Though I'm not aware of any other outstanding issues like this.

davezuko:
@nhagen I totally agree with it being optimized for production out of the box, so I'll keep playing around with solutions for this, but documentation will unfortunately have to suffice in the meantime.

PR #219, opened 2015-12-05
use `validVendor`

crossjs:
set `vendor_dependencies` with `validVendor`


davezuko:
Ah thank you, total lapse on my part. Doh.

PR #223, opened 2015-12-06
Improve createReducer

Mistereo:
This change allows us to write reducers which depend on additional data as described here: https://github.com/rackt/redux/issues/749#issuecomment-141570236

### DIFF
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -11,6 +11,7 @@ export function createConstants (...constants) {
 }
 
 export function createReducer (initialState, fnMap) {
+  // possible use of rest arguments: https://github.com/rackt/redux/issues/749#issuecomment-141570236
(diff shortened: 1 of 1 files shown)


davezuko:
Interesting. Would you mind adding a comment to link to that discussion above `createReducer` so people looking at the function understand the purpose of `...rest`?

Mistereo:
I've added comment, let me know if I need to correct the text (I'm not a native speaker).
I also wonder why we pass to reducer only `payload` from action?
If it's [FSA](https://github.com/acdlite/flux-standard-action) I think there can be cases when we want to access `meta` or/and `error` properties.

davezuko:
@Mistereo it was originally just a utility for my own use, and I only used it when it made sense. That said, I'll probably be changing it so it passes the entire action rather than just the payload.

PR #230, opened 2015-12-07
feat(cssModules): add loader configuration

davezuko:


davezuko:
@neverfox and @mtsr does this look right to you guys?

mtsr:
This works for me, but @neverfox suggested adding a separate loader for css-modules. I think splitting between the `styles/` and `components/` / `layouts/` / `views/` could work. I'm not sure about the implications of `@import`ing a non-css-modules from a css-module, though.

mtsr:
I just tested with `url('./some.png')`, and that fails (after I add the png url-loader, of course). I feel it's odd, since the scss entry point is in the same folder as the image. I think this is something worth fixing, as it will be surprising to users.

mtsr:
:global {} is something I still need to test. I had some issues with foundation-sites before, which I'm curious if they're present now as well.

davezuko:
I haven't played with `url(...)`, but is there a reason why that wouldn't work given the current implementation? I don't see anything in the webpack configuration that would conflict with this, so is it just that additional setup is required?

mtsr:
Nope, AFAICT url’s should be resolved relative to the entry point. But apparently the index.html or something at that level is being considered the entry point, instead of the component/layout/view scss file.

davezuko:
@mtsr I've managed to get images to load correctly in this branch (just created `foo.png` in `views`) and then added:

```
:global {
  body {
    background: url('./foo.png');
  }
}
```

Curious to see if this is fixed for you or if you're still having problems.

davezuko:
Going to merge this since I have some related changes to make. We can continue to debug any issues in additional PR's.

mtsr:
I made some changes to my own project to exclusively apply css-modules to `/src/(components|layouts|views)`, and use the regular pipe-line for `/src/styles`. This is working well for us, although I haven't yet tested what happens when I `@import` in a css-module from the non-modules css.

Is this something that is of interest?

davezuko:
@mtsr I could see that being a nice option that can be toggled via the config file. I think it might confuse people as a default, but it does sound very practical.

mtsr:
The main reason to do this was some kind of parsing issue when wrapping `@import ‘foundation’; @include foundation-everything;` in `:global {}`.

I’ll make a PR this weekend or next week.

PR #231, opened 2015-12-07
refactor(config): use POJO rather than ES6 map

davezuko:


mtsr:
I think I actually prefer this syntax anyway.

gaboesquivel:
:+1:

PR #246, opened 2015-12-10
fix(karma): Avoid run tests twice when autoWatch is set to true

Novtopro:
I found that the tests will run twice when file changes occurs when the Karma is set to autoWatch. For more information, please check this out: https://github.com/nikku/karma-browserify/issues/67.

### DIFF
--- a/build/karma.js
+++ b/build/karma.js
@@ -4,20 +4,22 @@ import config   from '../config';
 const debug = require('debug')('kit:karma');
 debug('Create configuration.');
 
-const KARMA_ENTRY_FILE = 'karma.entry.js';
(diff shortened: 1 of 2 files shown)


davezuko:
Hey thanks, we have a pretty big PR coming for Babel 6 so I'd like to wait for that to be merged before pulling this in just to make that process easier.

Novtopro:
:ok_hand:

davezuko:
Works great, thank you for the fix!

BebeSparkelSparkel:
Seems like this is still a thing.
I'm getting it and there is an article on it as well
https://pracucci.com/avoid-karma-tests-run-twice-on-autowatch.html

PR #258, opened 2015-12-11
refactor(utils): break out utility functions

justingreenberg:
trying to make the utilities more self documenting—i keep finding myself parsing through `utils/index`. 

since they are utilities and will be used daily but never looked at, they include both named as well as default exports. this should make them pretty mistake-proof

```js
import { createConstants } from 'utils/createConstants';
/*  or  */
import createConstants from 'utils/createConstants';
```

i also left the index.js entry to support imports via the `utils` webpack alias, ie:

```js
import { createConstants } from 'utils/';
/*  or  */
import createConstants from 'utils';
```

which is a nice syntax and makes the update 100% backwards-compatible with the previous implementation, but is definitely not necessary... let me know what you think

by the way, is it just me or did the builds get SERIOUSLY faster?


davezuko:
Was actually planning on this ^^, thank you. And I haven't really noticed, maybe I'll do a comparison of build times later.

rsilvestre:
500th commits. Congratulation !

davezuko:
@rsilvestre :tada: @ @justingreenberg 

_(so many @ symbols)_

PR #259, opened 2015-12-11
docs(readme): add styles folder

inooid:
I was copy pasting some README elements to my own app's README and figured that the `styles` folder is left out in the *structures* part.


davezuko:
Good catch, thank you!

PR #262, opened 2015-12-11
feat(redux): add ...rest to createReducer

justingreenberg:
@Mistereo @davezuko `...rest` param bug fixed in #258

note: destructuring the `action` object in the returned function args does not work. i tried using default params with same results... assignment moved inside function body `const { type, payload } = action;`


davezuko:
My github skills are lacking, rebased this locally to try to get rid of the stupid merge commits but the PR tool doesn't pick up on the merge. This is in master!

PR #271, opened 2015-12-12
feat(standard): add style config and refactor to match

davezuko:
@justingreenberg @mtsr thoughts?

@justingreenberg I found out that standard-style is actually fairly close to what I had, save for my obsessive import and colon alignment (which should be dropped anyways to improve git diffs). Also, standard style enforces `function () {` instead of the `function() {` from rackt -- which I will fight to the death.

I know it's a big change, and it may alienate some people, but that's why I'm not just merging my own PR :).

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,10 +5,13 @@ React Redux Starter Kit
 [![Build Status](https://travis-ci.org/davezuko/react-redux-starter-kit.svg?branch=master)](https://travis-ci.org/davezuko/react-redux-starter-kit?branch=master)
 [![dependencies](https://david-dm.org/davezuko/react-redux-starter-kit.svg)](https://david-dm.org/davezuko/react-redux-starter-kit)
 [![devDependency Status](https://david-dm.org/davezuko/react-redux-starter-kit/dev-status.svg)](https://david-dm.org/davezuko/react-redux-starter-kit#info=devDependencies)
-
+[![js-standard-style](https://img.shields.io/badge/code%20style-standard-brightgreen.svg)](http://standardjs.com/)
 
 > ### This Project Recently Upgraded to Babel 6!
 > Woohoo! If you'd like to try it out, you're welcome to build directly from the master branch. However, if troubleshooting issues with Babel isn't quite your thing, just pull the [stable v0.18.0 release](https://github.com/davezuko/react-redux-starter-kit/tree/v0.18.0) and continue on your way with Babel 5.
+
+> ### Want Semicolons?
+> After installing npm dependencies, open `.eslintrc` change the `semi` rule from `never` to `always` and then run `npm run lint:fix`. Easy as that! Or use the same npm script after installing your preferred ESLint configuration; it's easy to customize the project's code style to suit your team's needs.
 
 This starter kit is designed to get you up and running with a bunch of awesome new front-end technologies, all on top of a configurable, feature-rich webpack build system that's already setup to provide hot reloading, CSS modules with Sass support, unit testing, code coverage reports, bundle splitting, and a whole lot more.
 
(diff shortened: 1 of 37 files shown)


davezuko:
Also pinging @patrickheeney

patrickheeney:
Major :+1: I painstakingly adjusted the imports and colons on every file the first thing after installing this repo. Only difference on my end now is the `;`. 

> The primary goal of this project is to remain as unopinionated as possible. 

Definitely true to the project goals ^.

davezuko:
@patrickheeney good point, I realized how annoying it could be for other people, and usability/functionality definitely should come before "looking pretty". So while the spacing changes definitely reflect the project goal of _as unopinionated as possible_, what are your thoughts as to how the no-semis rule plays into that?

patrickheeney:
I personally am in favor of adding `;` and `function() {}` with the extra space. I haven't researched much into the styles though to know where *I* fit. Is there any way to measure the popularity of different style guides?

README.md
```diff
@@ -5,11 +5,14 @@ React Redux Starter Kit
 [![Build Status](https://travis-ci.org/davezuko/react-redux-starter-kit.svg?branch=master)](https://travis-ci.org/davezuko/react-redux-starter-kit?branch=master)
 [![dependencies](https://david-dm.org/davezuko/react-redux-starter-kit.svg)](https://david-dm.org/davezuko/react-redux-starter-kit)
 [![devDependency Status](https://david-dm.org/davezuko/react-redux-starter-kit/dev-status.svg)](https://david-dm.org/davezuko/react-redux-starter-kit#info=devDependencies)
-
+[![js-standard-style](https://img.shields.io/badge/code%20style-standard-brightgreen.svg)](http://standardjs.com/)
 
 > ### This Project Recently Upgraded to Babel 6!
 > Woohoo! If you'd like to try it out, you're welcome to build directly from the master branch. However, if troubleshooting issues with Babel isn't quite your thing, just pull the [stable v0.18.0 release](https://github.com/davezuko/react-redux-starter-kit/tree/v0.18.0) and continue on your way with Babel 5.
 
+> ### Want Semicolons?
```
davezuko: @patrickheeney Added a note here about how to re-add semi colons for those who want them. It's incredibly easy.

justingreenberg:
@davezuko this looks good to me! :+1:  especially thank you for dropping colon alignment i know that was probably a hard pill to swallow... with respect to semicolons, losing them gave me pause for a second but since you documented reverting and it's is so easy i don't think it's an issue at all

davezuko:
Cool, thanks @justingreenberg. You are pro no-semis though, right? That is at least similar to the rackt config.

davezuko:
Going to merge this so I can get to work on other features. I think regardless of the semi vs. no-semi debate, it's good to get a lot of these changes in the starter kit so it's easier to work with.

justingreenberg:
absolutely, people just need to be cognizant of [**automatic semicolon insertion**](http://bclary.com/2004/11/07/#a-7.9.1) at runtime (maybe link to spec?)... the only caveat that seems to cause issues is splitting `return`, `throw` or `break` statements onto multiple lines, since asi will kick in and next lines will not execute 

since this is a big change, one other thing that strikes me as off is hardcoding the `kit:` namespace into each debug instantiation... that's not really how the library was designed and i think it's unnecessary anyway since this is just a starter

patrickheeney:
I am happy with this. Great work!

sarunast:
For me the no semicolon style was quite new I didn't even knew that you can write code that way. I've been trying to use this style for the last 2 days and my brains starting to 'understand' it :smile:  but it is very strange when I've never seen code like this before. It would be great if someone could explain why I I should stop using semicolons and what benefits it offers, any links are welcome too.

By the way there was recently a survey for JS developers more then 5000 devs participated. What is interesting only 11% doesn't use semicolons and 80% use semicolons. So I guess it is really worth thinking what default style should be.
https://ponyfoo.com/articles/javascript-developer-survey-results

patrickheeney:
I personally am in favor of semicolons. However I think at this point is mostly user preference, like tabs vs spaces. There are other things to consider if you really wanted to get into development methodology. At the end of the day there are so many things to build, use whatever doesn't slow you down. https://github.com/davezuko/react-redux-starter-kit#want-semicolons and a few seconds later you are back in action!

Issue #273, opened 2015-12-12
DEBUG=kit:* enabled by default; opinions


davezuko:
@justingreenberg brought this up in https://github.com/davezuko/react-redux-starter-kit/pull/271. For an explanation, I enabled it because I wanted users to know what was happening with webpack, especially for things like the development server (i.e. where is it running?). I understand that the [debug](https://www.npmjs.com/package/debug) package was designed so that debug output would be disabled by default and selectively enabled, so maybe a good compromise would be to do one of the following:

1) Use `console.log` for output that we want to always be shown (`Server running at...`).

2) Use a separate namespace for those types of outputs ^^, and configure the `DEBUG` environment variable to only show these by default. This still strays slightly from its intended usage, but I think defaulting to more feedback is better than not, especially when it's easy to disable by editing the `.env` file.

There are always options 3 and 4: disable it entirely by default, or keep it as is.

Personally, I like seeing all the output for the build pipeline (which is probably obvious seeing as I'm the maintainer, -_-), so I slightly disagree with disabling the debug output, but it's open for discussion.

justingreenberg:
i like seeing all output for build pipeline too! it's nice and colorful and makes me happy. i agree with keeping all output and just dropping the `kit:` prefix. 

```
  kit:config Create configuration. +0ms
  kit:config Apply environment overrides. +16ms
  kit:webpack:_base Create configuration. +935ms
  kit:webpack:development Create configuration. +11ms
  kit:webpack:development Override devtool with cheap-module-eval-source-map. +0ms
  kit:webpack:development Enable Hot Module Replacement (HMR). +0ms
  kit:webpack:development Apply react-transform-hmr to babel development transforms +6ms
  kit:server:webpack-dev Enable Webpack dev middleware. +68ms
  kit:server:webpack-hmr Enable Webpack Hot Module Replacement (HMR). +28ms
  kit:bin:server Server is now running at localhost:3000. +160ms
```

becomes:

```
  config Create configuration. +0ms
  config Apply environment overrides. +16ms
  webpack:_base Create configuration. +935ms
  webpack:development Create configuration. +11ms
  webpack:development Override devtool with cheap-module-eval-source-map. +0ms
  webpack:development Enable Hot Module Replacement (HMR). +0ms
  webpack:development Apply react-transform-hmr to babel development transforms +6ms
  server:webpack-dev Enable Webpack dev middleware. +68ms
[...]

PR #274, opened 2015-12-12
chore(debug): drop kit:* prefix

davezuko:
fix(debug): drop kit: prefixes in server code


davezuko:
See https://github.com/davezuko/react-redux-starter-kit/issues/273#issuecomment-164176989

Issue #277, opened 2015-12-13
Hot Module Reloading not working with functional components


ntgn81:
Hot Module Reloading is not yet working with functional components, according to  https://github.com/gaearon/babel-plugin-react-transform/issues/57

Took me some time to get to that link. Just putting this out here in case someone else has the same issue with `AboutView` not hot reloading.

Looks like until it's fixed, it might be better to not use functional components. HMR gives me so much more than functional components at this point.

Also, this is a great starter kit. Exactly what I wanted without all the universal fuzz, making it so much harder to learn. Thank you for the great work! :star2:

mtsr:
Thanks for sharing, good to know why my About Page wasn’t reloading.

justingreenberg:
@mtsr @ntgn81 see discussion on #279 for explanation and discussion... 

TL;DR: react-transform does not support functional components. this is known, documented and deliberate—this issue can be closed

@davezuko it might be a good idea to add a note to docs about this issue

davezuko:
@justingreenberg My vote is that we just make everything extend `React.Component` for now until there's some movement on this from react-transform-hmr. I'd like to avoid people submitting bugs or thinking that the starter kit doesn't work because they're messing with pre-packaged stateless components. A note would still be nice to add so people who incorporate their own know why HMR doesn't work for those.

patrickheeney:
@davezuko Sounds good. Are there any optimizations for production in place yet? For example the one mentioned with #279.

justingreenberg:
they are still considered pure components as long as they only have a render method. @patrickheeney i have a [working branch using `babel-plugin-react-pure-components`](https://github.com/davezuko/react-redux-starter-kit/compare/master...justingreenberg:babel-pure-components) (written by babel core contributor) that detects pure components and transforms them into function components for optimized production builds.

i haven't opened a pr yet because i'm using a local `npm link`ed version of the plugin... the published npm package contained deps that were out of date and must be updated prior to merge, since it's not working "out of the box" yet

i dug a little deeper and it looks like referentially transparent components **in between** root and edge nodes are no longer an issue since the rewrite of react-transform... transforms are not applied to function components but the plugin continue to traverse it's children. this means that the Root component must be component class, however something structural such as the CoreLayout component could be a function component (since it doesn't "need" HMR) and transforms still work on HomeView. 

This is essentially what I did in #292, with a explanation

mtsr:
[...]

PR #279, opened 2015-12-13
Convert classes to stateless components

patrickheeney:
It might be worthwhile to create a view to demonstrate a state-full component, maybe a button that toggles something.

### DIFF
--- a/src/containers/Root.js
+++ b/src/containers/Root.js
@@ -2,37 +2,40 @@ import React from 'react'
 import { Provider } from 'react-redux'
 import { Router } from 'react-router'
 
-export default class Root extends React.Component {
-  static propTypes = {
-    history: React.PropTypes.object.isRequired,
-    routes: React.PropTypes.element.isRequired,
-    store: React.PropTypes.object.isRequired
-  }
-
-  render () {
-    const content = (
-      <Router history={this.props.history}>
-        {this.props.routes}
-      </Router>
-    )
-
-    if (__DEBUG__ && !__DEBUG_NW__) {
-      const DevTools = require('containers/DevTools').default
-
-      return (
-        <Provider store={this.props.store}>
-          <div>
-            {content}
-            <DevTools />
-          </div>
-        </Provider>
-      )
-    } else {
-      return (
-        <Provider store={this.props.store}>
-          {content}
-        </Provider>
-      )
-    }
-  }
+const PropTypes = {
+  history: React.PropTypes.object.isRequired,
+  routes: React.PropTypes.array.isRequired,
+  store: React.PropTypes.object.isRequired
 }
+
+const RootContent = ({ history, routes }) => {
+  return (
+    <Router history={history}>
+      {routes}
+    </Router>
+  )
+}
+
+const Root = ({ history, routes, store }) => {
+  if (__DEBUG__ && !__DEBUG_NW__) {
+    const DevTools = require('./DevTools').default
+    return (
+      <Provider store={store}>
+        <div>
+          <RootContent history={history} routes={routes} />
+          <DevTools />
+        </div>
+      </Provider>
+    )
+  }
+
+  return (
+    <Provider store={store}>
+      <RootContent history={history} routes={routes} />
+    </Provider>
+  )
+}
+
+Root.propTypes = PropTypes
+
+export default Root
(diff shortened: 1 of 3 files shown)


justingreenberg:
@patrickheeney have you tested this with HMR? react-transform does not currently support functional components...

patrickheeney:
@justingreenberg It is unfortunate, but the only time I generally update stateless components is to edit the html, at which point I refresh. The modifications to the state or other parts of the app still work with HMR. How often do you edit CoreLayout or Root container? Hopefully they fix upstream soon like the babel 6 issue.

Related #277.

justingreenberg:
@patrickheeney this breaks HMR for HomeView and other presentational components as well

patrickheeney:
@justingreenberg Indeed, just the home view, but I would argue the others are rarely changed. However, this coincides with the existing AboutView so the current behavior of refreshing these two views would be similar.

justingreenberg:
@patrickheeney yea I actually [added AboutView as functional component](https://github.com/davezuko/react-redux-starter-kit/commit/db66626ca8a02ecf030a3f7f5a669ac338fd5897) in my PR migrating to `redux-simple-router`, but it was really just meant to demo routing behavior

personally, i leverage HMR to maintain component state when building, styling and testing UI. this 
would break anything attached to the stateless components (styles, classes, etc)...
 
also the issue isn't babel 6, it's actually react itself—currently there is no good way to track changes to these components since they are just simple functions that render html, whereas class components are tracked using their displayName. one of the babel authors outlines the problems better than i could in gaearon/babel-plugin-react-transform#57

patrickheeney:
I will close this now as the starter kit is likely to change by the time this is no longer an issue. Secondly I didn't update the tests. I was trying to say the issue is similar to babel 6, in that we are essentially waiting.. A "bug" in an upstream library prevents the usage of class based components with super(), and in this case a library preventing stateless components.

justingreenberg:
another solution would be to use *pure* component classes (*only* render method) during development to support HMR development workflow, then use the babel pipeline to transform them into pure functions for production using [babel-plugin-react-pure-components](https://github.com/thejameskyle/babel-plugin-react-pure-components)

patrickheeney:
@justingreenberg Interesting. 

Also, this worked for me as another idea. What about you?

```
class HomeViewWrapper extends React.Component {
  render () {
    return <HomeView {...this.props} />
  }
}

HomeViewWrapper.propTypes = PropTypes

export default connect(mapStateToProps, counterActions)(HomeViewWrapper)
```

patrickheeney:
Just thinking out loud. The wrapper could be useful when developing a component, when you are done with it, comment it out for the pure component. I also like the idea of using the babel plugin for production so I might explore that as well.

justingreenberg:
@patrickheeney yea wrapping with class is a good pattern for adding lifecycle methods and fixtures for testing (and yes would enable HMR) but this wouldn't really add much practical value over pure component classes at present. it should be noted that the react team is actively working on the referential issue and and optimizations of functional components. 

i've been using [recompose](https://github.com/acdlite/recompose) a lot lately, which is an awesome toolkit for composing pure components—when referential support is release it is going to be a game-changer!

patrickheeney:
@justingreenberg I thought of it merely as a means to enable HMR for development. Basically you would uncomment the wrapper to dev a component, re-comment when it is "production ready". For example CoreLayout for me will never change, so I would likely always leave it a pure component. If I need to edit it where a refresh wouldn't do the job, I would wrap it for HMR. (Only temporarily until the issue is fixed).

I will have a look at recompose, sounds pretty cool.

justingreenberg:
yeah i hear you. the problem is react-transform needs references to every upstream component to find child components. so if you wanted to enable HMR for, let's say HomeView, you would have to wrap Root and CoreLayout as well as HomeView itself. obviously this would become extremely tedious as the component tree grows...

patrickheeney:
I didn't do any extensive testing but I only wrapped home view, the other
components were functions and HMR worked as expected. Since stateless never
change between refreshes you would have to wrap any component you want to
edit between those refreshes. If your not editing core layout for example
it should be OK. At least that is my take.

Anyway, will leave this tabled for now and see how things progress
upstream.

gaearon:
You might be interested in https://github.com/reactjs/redux/pull/1455.

gaearon:
React Hot Loader 3 supports functional components without destroying the state.
It supersedes both React Hot Loader and React Transform, and is built on lessons learned from both.
Check it out: https://github.com/gaearon/react-hot-boilerplate/pull/61

PR #280, opened 2015-12-13
Only emit warning from eslint-loader. Fixes #278

### LINKED ISSUE #278: eslint errors kill index.html bundling
It seems that any eslint error now prevents index.html from from ending up in the bundled files and you get a resource not found error (from express?). To recreate, just add a semicolon at the end of a line anywhere, run or restart dev, and you'll get a `Cannot GET /` error in the browser (and you won't see index.html in the bundle asset list).

If this was intentional, I think it goes too far. While eslint errors should paint red all over your console so you'll fix it, it shouldn't stop a valid codebase from running (as it didn't before). For one, it blocks quick and dirty things like throwing in a quick debugger statement.

If it was unintentional, any ideas on how to fix it so things don't just blow up? Might this be a side-effect of switching to express with middleware over the webpack-dev-server? (Sorry for being a bit behind on catching up with recent changes).

neverfox:


davezuko:
Awesome, might end up only enabling this when HMR is enabled so that development builds can be failed, but this works for now. Thanks!

neverfox:
Is that not the case here since it's inside of the `if` for `config.compiler_enable_hmr`?

davezuko:
Oh whoops, my mistake -- you're absolutely right. Didn't look too closely during my flurry of merges.

PR #281, opened 2015-12-13
Feature/universal

ywarezk:
Added Server Side Rendering to your starter kit using hapi. 
you can run it by: `npm start production`


davezuko:
Hey, I really appreciate all of the work you did for this, but unfortunately one of the main goals of this starter kit is to avoid complexity such as this. A couple of months back this project did actually ship with support for rudimentary isomorphic rendering, but that caused fairly significant feature and scope creep. Ultimately, it was decided that we keep our focus more narrow and allow people to extend the repo as it suits them.

That said, it _would_ be awesome to get some documentation in place to explain how one might go about implementing isomorphic rendering, but I don't think we'll ever go back to shipping it out of the box. Thanks again for your work, it's appreciated.

mtsr:
I guess in theory one could keep a branch that has universal rendering, either as an example or even fully fledged. Although I can see why you might not want to maintain that branch personally.

ads1018:
+1 Getting some documentation in place to explain how one might go about implementing universal rendering :)

davezuko:
@mtsr I've tried that before but it just gets so incredibly difficult to maintain.

mtsr:
I know. But maybe someone else who wants universal rendering for this starter kit is willing to maintain it?

PR #283, opened 2015-12-13
fix(babel): fix export interop, remove hotfix

justingreenberg:
i'm not sure when or how the plugin broke but this fixes es6 default export interop with commonjs and removes the `{ default }` hotfix for inline require statements


davezuko:
Thanks as always!

PR #285, opened 2015-12-13
refactor(debug): update kit namespace to app

### LINKED ISSUE #273: DEBUG=kit:* enabled by default; opinions

justingreenberg:
updated #274, per discussion—resolves #273

### DIFF
--- a/.env
+++ b/.env
@@ -1,2 +1,2 @@
 NODE_ENV=development
-DEBUG=kit:*
+DEBUG=app:*
--- a/README.md
+++ b/README.md
@@ -93,7 +93,7 @@ Before delving into the descriptions for each available npm script, here's a bri
 * Compiling the application to disk? Use `npm run compile`.
 * Deploying to an environment? `npm run deploy` can help with that.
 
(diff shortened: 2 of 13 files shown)


davezuko:
Thank you! :beers:

PR #292, opened 2015-12-14
Pure component notes

### LINKED ISSUE #277: Hot Module Reloading not working with functional components

justingreenberg:
resolves #277

### DIFF
--- a/tests/layouts/CoreLayout.spec.js
+++ b/tests/layouts/CoreLayout.spec.js
@@ -34,11 +34,4 @@ describe('(Layout) Core', function () {
     expect(_component.type).to.equal('div')
   })
 
-  it('Should render a child component.', function () {
-    const child = TestUtils.findRenderedDOMComponentWithClass(
-      _rendered, 'child'
-    )
-
-    expect(child).to.exist
-  })
 })
--- a/src/layouts/CoreLayout.js
+++ b/src/layouts/CoreLayout.js
@@ -1,18 +1,27 @@
 import React from 'react'
 import 'styles/core.scss'
 
-export default class CoreLayout extends React.Component {
-  static propTypes = {
-    children: React.PropTypes.element
-  }
-
-  render () {
-    return (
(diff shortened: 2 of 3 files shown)


tests/layouts/CoreLayout.spec.js
```diff
@@ -34,11 +34,4 @@ describe('(Layout) Core', function () {
     expect(_component.type).to.equal('div')
   })
 
-  it('Should render a child component.', function () {
```
davezuko: This is because the test utils don't work with pure components yet, right?

justingreenberg: exactly... do you want me to revert this commit to leave this as component class?

davezuko: Not at all, was just making sure. And I'm not totally caught up to speed here: A regular React component class rendered inside of a pure component will still work w/ HMR, right?

justingreenberg: @davezuko yes all transforms are working, see below as usual. i also tested hmr with multiple nested `AboutView`s to be sure react-transform is walking the full component tree :+1: 

hmr-stateless

justingreenberg: on a separate note, do you want me to make a nice demo screencast for readme?

davezuko: That would be _amazing_, and thanks for all the gifs, they help a lot!

PR #293, opened 2015-12-14
Implement actions and reducers with redux-actions

neverfox:
Also, gets action type  in compliance with duck-modular-redux:

```
MUST have action types in the form npm-module-or-app/reducer/ACTION_TYPE
```


davezuko:
Sorry to split the convo, but I posted a comment here: https://github.com/davezuko/react-redux-starter-kit/issues/204#issuecomment-164334099

neverfox:
I compromised on the action type requirement, if you prefer that. If you want me to go all the way back to no namespace, I can make that change.

src/redux/modules/counter.js
```diff
@@ -1,17 +1,15 @@
-import createReducer from 'utils/createReducer'
+import { createAction,
```
davezuko: Not at all relevant to the PR, but just curious: is this a coding convention you use in your projects for imports? Never seen it before (not saying to change it, just curious). Looks kinda cool.

davezuko:
:+1: Awesome

src/redux/modules/counter.js
neverfox: You mean having mutiple named imports on different lines? Yes, I do, because the lists can get to be long in some cases, so I just adopt the rule of new-lining more than one, though I'm not always consistent when it's just two.

davezuko: I just meant the indentation of it, since normally I see:

```
import {
  foo,
  bar
} from 'baz'
```

mtsr:
One thing I'm missing is the initial state. Where does that get defined now?

PR #294, opened 2015-12-14
Improve Root

neverfox:
Use getters in Root, add 100% height up to start of app content (the developer can take it from there), and some additional best-practice starter CSS rules.

### DIFF
--- a/src/containers/Root.js
+++ b/src/containers/Root.js
@@ -9,30 +9,29 @@ export default class Root extends React.Component {
     store: React.PropTypes.object.isRequired
   }
 
-  render () {
-    const content = (
+  get content () {
+    return (
       <Router history={this.props.history}>
         {this.props.routes}
       </Router>
     )
+  }
 
+  get devTools () {
     if (__DEBUG__ && !__DEBUG_NEW_WINDOW__) {
       const DevTools = require('containers/DevTools')
-
-      return (
-        <Provider store={this.props.store}>
-          <div>
-            {content}
-            <DevTools />
-          </div>
-        </Provider>
-      )
-    } else {
-      return (
-        <Provider store={this.props.store}>
-          {content}
-        </Provider>
-      )
+      return <DevTools />
     }
+  }
+
+  render () {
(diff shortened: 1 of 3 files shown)


davezuko:
Very cool, I like it a lot. What's the benefit of adding the `height:100%` styles?

davezuko:
Also, part of the reason for the `if/else` was to solve this issue: https://github.com/davezuko/react-redux-starter-kit/issues/189

Any thoughts on that?

neverfox:
I'm assuming/hoping that because the getter is still conditional, returning null otherwise, it will have the same dead code effect. What's the way you were checking for that?

As for the 100%, see [here](http://mazharahmed.me/get-100-height-working-in-css/). Setting it also on the div in Root ensures that that rule carries down through all the React gunk we have to put between the  root div in index.html and the first real app content (where most developers will start writing their own CSS styles).

neverfox:
Most of those rules are lifted from here: https://github.com/cssrecipes/defaults

davezuko:
This looks like it would work just fine, the only thing is in production/non-debug mode:

```
+  render () {
+    return (
+      <Provider store={this.props.store}>
+        <div style={{ height: '100%' }}>
+          {this.content}
+          {this.devTools}
+        </div>
+      </Provider>
+    )
+  }
```

Since devtools is null, you'd have an extra wrapping div that serves no purpose. Honestly that's not that big of a deal, and this refactoring makes the code look _much_ cleaner so I won't lose any sleep over an extraneous div; the bigger problem was with the devtools imports still existing in prod builds (which has been solved by using `__DEBUG__` directly rather than the previous React props).

neverfox:
Actually, my intent is for the wrapping div to be there, even without devtools, in order to give a 100% height parent for all user components and views to work from, so it's doing double duty here. Honestly, you could just as well put devtools outside of it.

davezuko:
As usual I'm just dumb and totally ignored the effect the Provider would have.

Looks great, thank you. :sparkles:

PR #297, opened 2015-12-14
fix(app): fix root element style so that height 100% applies

neverfox:
For some reason, the root class didn't "take" and the style approach feels cleaner since it's out of the way.

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -7,6 +7,6 @@
   <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.6/css/bootstrap.min.css">
 </head>
 <body>
-  <div id="root" class="root"></div>
+  <div id="root" style="height: 100%"></div>
 </body>
 </html>
--- a/src/styles/_base.scss
+++ b/src/styles/_base.scss
@@ -17,8 +17,4 @@ body {
   box-sizing: inherit;
 }
 
-.root {
-  height: 100%;
-}
-
 // Additional settings go here


davezuko:
Thanks @neverfox, looks good. One thing I didn't catch in the PR to add the styles was that they were added in `_base.scss` rather than `core.scss`. Could you shift all of the styles into core with this PR? Kill two birds with one stone, since `_base.scss` shouldn't include any actual styles (I need to document that better).

neverfox:
Oh? Ha. I've been using _base.scss for my styles (what little global ones I have). I was wondering why it was placed before normalize. Would you rather though that core.scss stay "pure" as an import consolidator? If so, I could make a _defaults.scss file.

neverfox:
While I'm at it, how do you feel about moving normalize into a dependency with `normalize-scss` (now that it no longer requires compass) so the kit can stop hauling around that vendor file?

neverfox:
What kind of things did you have in mind for _base.scss?

davezuko:
`_base.scss` is a pattern I was introduced to where it acts as a global import (similar to what your `_defaults.scss` sounds like?) and as a result has to remain pure of any actual styles. It would just be variables, mixins, and whatever else you might need. I'm not sure this is even necessary anymore, it's been a while since I've done any real sass work so the variables/etc. might just be available globally.

As far as normalize-scss, possibly, but don't throw that one in just yet. Keeping it might be helpful for the time being in case people don't want to use sass but still need to import normalize (only counter argument I can think of at the moment). I'll probably change my mind by morning though, hah.

neverfox:
Then what I would suggest is just running with it as is since it does align with most of the current organization suggestions:

core.scss (serves purely as a manifest to bring together all partials, import this into JS)
_base.scss (default global styles not purely related to design, basically enhancements to normalize)

from there people can add other partial files and add them to core, e.g. _variables.scss, _theme.scss.

If we go that way, then I would move the base import after the normalize import, since normalize should be the tabula rasa everything else works from.

I'll let you sleep on it and I can make any changes tomorrow.

davezuko:
So I'm in agreement, but maybe I should clarify `_base` a bit further. `_base` is to `_variables_`, `_theme`, etc. as `core.scss` is to all of style partials. So `_base` might look like:

```
import variables from './variables'
import theme from './theme' // where these are all just overrides for styles
// etc.
```

And then `core` only has to bring in `base` and them be responsible for loading in all of the partials. From an old project, so not the greatest to showcase, but here's an example: https://github.com/davezuko/jackie/blob/master/client/sass/base/_base.scss

Anyways, happy to pick this up again in the morning.

neverfox:
Okay, that gives me the idea. I haven't had time to rework this yet though.

davezuko:
@neverfox if you want I can just merge this and make the adjustment myself.

PR #305, opened 2015-12-14
feat(server): Static Assets

patrickheeney:
I removed comments as express can be run as your web server, and in the development mode it already is doing that. It might be worthwhile to remove the debug() production notice as well as this could be a universal starter kit.

### DIFF
--- a/server/app.js
+++ b/server/app.js
@@ -28,11 +28,8 @@ if (config.compiler_enable_hmr) {
     'more about deployment strategies, check out the "deployment" section ' +
     'in the README.'
   )
-
-  // Serving ~/dist by default. Ideally these files should be served by
-  // the web server and not the app server, but this helps to demo the
-  // server in production.
-  app.use(express.static(paths.base(config.dir_dist)))
 }
+
+app.use(express.static(paths.base('static')))
 
 export default app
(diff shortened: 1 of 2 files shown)


davezuko:
Thanks! This could work, but I'm still curious to see what @justingreenberg is referring to with the file-loader for copying `~/src/assets` (or `~/src/static`) to the root of the dist folder. I haven't had much of a chance to look at the documentation for this, so I'll be holding off on this PR for a little bit.

davezuko:
@patrickheeney what is your opinion on following your PR (where the assets are in the root of the project) vs putting them in src and compiling to dist? The latter is how it's currently done, but if you have a good argument for keeping them in the root we could drop my code and go forward with your implementation.

davezuko:
Actually, one good argument for your solution is that it the assets are available during development, whereas with the current implementation they're not. Though, thinking back to why I was originally against the separate `static` directory in root is because then your application code is split up into two directories. Rawr.

patrickheeney:
I picked root static dir over src since it seemed related more towards server artifacts versus application resources, but there was no strong preference behind it. It sounds like this is now integrated, so closing this.

PR #308, opened 2015-12-14
fix(config): Make babel plugin overrides more flexible

### LINKED ISSUE #291: Webpack HMR plugin conflict
I have a custom loader for my application that is something like this:

```
{
  test: /custom.js$/,
  exclude: /node_modules/,
  loader: paths.client('lib/bundler'),
}
```

This conflicts with the HMR loader in `_development.js` because it checks for `/js/` in `loaders.test` then adds to this array `loader.query.env.development.plugins[0][1].transforms`. This conditional looks to be babel specific so instead of the regex test, what about just testing for the loader `if (loader.loader === 'babel')`?

patrickheeney:
To fix #291

### DIFF
--- a/build/webpack/_development.js
+++ b/build/webpack/_development.js
@@ -32,7 +32,6 @@ export default (webpackConfig) => {
     // but _only_ when HMR is enabled. Putting this in the default development
     // configuration will break other tasks such as test:unit because Webpack
     // HMR is not enabled there, and these transforms require it.
-    let appliedHmr = false
     webpackConfig.module.loaders = webpackConfig.module.loaders.map(loader => {
       if (loader.loader === 'babel') {
         debug('Apply react-transform-hmr to babel development transforms')
@@ -43,32 +42,17 @@ export default (webpackConfig) => {
           locals    : ['module']
         }
 
-        loader.query.env.development.plugins =
-          loader.query.env.development.plugins.map(plugin => {
-            let [name, options, ...rest] = plugin
-            if (name !== 'react-transform') {
-              return plugin;
-            }
+        if (!loader.query.env.development.plugins[0][0] !== 'react-transform') {
+          debug('ERROR: react-transform must be the first transform plugin')
+          return loader
+        }
 
-            appliedHmr = true
-            options = {
-              ...options,
-              transforms: [
-                ...options.transforms,
-                reactTransformHmr
-              ]
-            }
-
-            return [ name, options, ...rest ]
-          })
+        loader.query.env.development.plugins[0][1].transforms
+          .push(reactTransformHmr)
       }
 
       return loader
     })
-
-    if (!appliedHmr) {
-      debug('Babel plugin react-transform-hmr could not be applied')
-    }
   }
 
   return webpackConfig
--- a/build/webpack/_development.js
+++ b/build/webpack/_development.js
@@ -36,15 +36,15 @@ export default (webpackConfig) => {
       if (loader.loader === 'babel') {
         debug('Apply react-transform-hmr to babel development transforms')
 
+        if (!loader.query.env.development.plugins[0][0] !== 'react-transform') {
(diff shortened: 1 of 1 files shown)


build/webpack/_development.js
```diff
@@ -41,12 +42,33 @@ export default (webpackConfig) => {
           imports   : ['react'],
           locals    : ['module']
         }
-        loader.query.env.development.plugins[0][1].transforms
-          .push(reactTransformHmr)
+
+        loader.query.env.development.plugins =
+          loader.query.env.development.plugins.map(plugin => {
```
justingreenberg: i mean, i know it's an ugly mutation but this really seems like overkill

patrickheeney: Overkill in what way?

patrickheeney: If you are referring to the verbosity of it all, I don't see a way to be flexible and succinct. It is a starter-kit so the order of the plugins is going to change ads more are added or removed.

justingreenberg: i'm sorry i wrote that really quickly didn't mean for it to come off like that!

yea i just meant it's a verbose solution for something that was one line. i hate that the mutation is hard-coded the way that it is, but it's so easy to understand the former way, and so complicated the latter. 

i totally understand on flexibility, but i think simplicity trumps in this instance. i'll leave it up to @davezuko  

if we leave it hard-coded, i agree we should make it a convention that react-transform must the first plugin and add a comment in _base webpack config

build/webpack/_development.js
```diff
@@ -32,8 +32,9 @@ export default (webpackConfig) => {
     // but _only_ when HMR is enabled. Putting this in the default development
     // configuration will break other tasks such as test:unit because Webpack
     // HMR is not enabled there, and these transforms require it.
+    let appliedHmr = false
     webpackConfig.module.loaders = webpackConfig.module.loaders.map(loader => {
-      if (/js(?!on)/.test(loader.test)) {
+      if (loader.loader === 'babel') {
```
davezuko: I'd prefer we just update this loader test and keep the rest of the original logic for the time being. @justingreenberg @patrickheeney thoughts? I know it could cause some issues with extensibility, but we can't cleanly accommodate every use case while remaining declarative.

build/webpack/_development.js
justingreenberg: @patrickheeney either way, this is a very elegant solution... the new plugin syntax is insane!

build/webpack/_development.js
justingreenberg: :+1:

build/webpack/_development.js
davezuko: @justingreenberg definitely agree ^^, I hate the syntax, it makes configuration horrendously complicated.

build/webpack/_development.js
patrickheeney: Since its a simple change, feel free to adjust on your next commit. I will close this PR.

patrickheeney:
@davezuko @justingreenberg PR updated. Is this acceptable?

davezuko:
Much easier to read, thank you.

build/webpack/_development.js
```diff
@@ -33,14 +33,20 @@ export default (webpackConfig) => {
     // configuration will break other tasks such as test:unit because Webpack
     // HMR is not enabled there, and these transforms require it.
     webpackConfig.module.loaders = webpackConfig.module.loaders.map(loader => {
-      if (/js(?!on)/.test(loader.test)) {
+      if (loader.loader === 'babel') {
         debug('Apply react-transform-hmr to babel development transforms')
 
+        if (!loader.query.env.development.plugins[0][0] !== 'react-transform') {
```
justingreenberg: much better than a comment in `_base.js` :+1: thank you!

PR #309, opened 2015-12-14
fix(config): Make css extract more flexible

### LINKED ISSUE #291: Webpack HMR plugin conflict

patrickheeney:
To fix #291

### DIFF
--- a/build/webpack/_production.js
+++ b/build/webpack/_production.js
@@ -18,7 +18,7 @@ export default (webpackConfig) => {
   debug('Apply ExtractTextPlugin to CSS loaders.')
   webpackConfig.module.loaders = webpackConfig.module.loaders.map(loader => {
     if (!loader.loaders ||
-        !loader.loaders.find(loader => /css/.test(loader.split('?')[0]))) {
+        !loader.loaders.find(name => /css/.test(name.split('?')[0]))) {
       return loader
     }
 
--- a/build/webpack/_production.js
+++ b/build/webpack/_production.js
@@ -16,25 +16,11 @@ export default (webpackConfig) => {
   }
 
   debug('Apply ExtractTextPlugin to CSS loaders.')
-  let appliedExtractCSS = false
   webpackConfig.module.loaders = webpackConfig.module.loaders.map(loader => {
-    if (!loader.loaders) {
+    if (!loader.loaders ||
+        !loader.loaders.find(loader => /css/.test(loader.split('?')[0]))) {
       return loader
     }
-
-    let hasCSSLoader = false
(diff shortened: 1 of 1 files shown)


build/webpack/_production.js
```diff
@@ -16,17 +16,37 @@ export default (webpackConfig) => {
...
+  let appliedExtractCSS = false
   webpackConfig.module.loaders = webpackConfig.module.loaders.map(loader => {
-    if (/css/.test(loader.test)) {
-      const [first, ...rest] = loader.loaders
+    if (!loader.loaders) {
+      return loader
+    }
+
+    let hasCSSLoader = false
+    loader.loaders.forEach(styleLoader => {
```
davezuko: Hey, this could probably be a bit short (while also removing the initial `loader.loaders` check at the top):

```js
if (
  !loader.loaders ||
  loader.loaders.reduce((hasLoader, loader) => {
    return hasLoader || /css/.test(loader.split('?')[0])
  }, false)
) {
  return
}
```

Thoughts?

davezuko: /cc @justingreenberg

patrickheeney: I personally am not fond of the inline function in the if statement, but if thats what you want? This could produce false positives with any loader with css in the name like `postcss` right?

davezuko: Good point, but what you ever use `postcss` without `css-loader`? I only mention the `css` shorthand since I was actually planning on removing the `-loader` suffixes.

And as far as the inline function, I'm just trying to condense the logic to keep the configuration a bit smaller. Honestly not that big of a deal. I think it's nice to avoid flag variables as much as possible, so this was following along those lines.

justingreenberg: you could add anchor operators to the regex ie `/^css$/.test(loader.split('?')[0])`

> I only mention the css shorthand since I was actually planning on removing the -loader suffixes.
:+1: the `-loader` is unnecessary

patrickheeney:
@davezuko @justingreenberg PR updated.

davezuko:
I like this a lot more, thank you ^^. I'll pull it down and rebase when I get a chance.

PR #312, opened 2015-12-15
chore(babel): update dependencies

### LINKED ISSUE #256: Error when calling super()
After updating my starter kit to match, I am getting this error in production when running npm run start. It seems to be caused by my React class calling the super constructor. The base starter kit doesn't seem to call super yet, but you will definitely run into this error after adding a few React components on top of the base. 

```
return obj && typeof _symbol2.default !== "undefined" && obj.constructor === _symbol2.default ? "symbol" : typeof obj === "undefined" ? "undefined" : (0, _typeof3.default)(obj);
```

Seemed to have tracked it here: https://phabricator.babeljs.io/T6670 and here: https://phabricator.babeljs.io/T6644. The cause seems to be updating the babel-runtime.

justingreenberg:
resolves #256, @neverfox @patrickheeney can you guys confirm?

https://phabricator.babeljs.io/T6644 resolved via https://phabricator.babeljs.io/rBCd2ee42194114e2350a0e8d55db78a74b720d8479

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,9 +38,9 @@
     "yargs": "^3.18.0"
   },
   "devDependencies": {
-    "babel-core": "^6.3.15",
-    "babel-eslint": "^5.0.0-beta4",
-    "babel-loader": "^6.1.0",
+    "babel-core": "^6.3.17",
+    "babel-eslint": "^5.0.0-beta6",
+    "babel-loader": "^6.2.0",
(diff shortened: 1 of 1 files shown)


neverfox:
:+1: Like a charm!

davezuko:
Awesome!

patrickheeney:
Just updated and works great. Thank you!

Issue #319, opened 2015-12-16
redux-dev-tools not removed when compiling for production env


ipanasenko:
Hey, guys!
I run following command: `NODE_ENV=production npm run compile` and I see redux-dev-tools code in `dist/app.**.js`

This command worked fine yesterday. Can you have a look, please?

davezuko:
I have an idea of what it might be, but absolutely no idea _why_. Checking.

ipanasenko:
I've created PR, ICYMI: https://github.com/davezuko/react-redux-starter-kit/pull/320

PR #320, opened 2015-12-16
fix(DevTools): devtools in compile for production

ipanasenko:
fixes https://github.com/davezuko/react-redux-starter-kit/issues/319


davezuko:
I give dead code removal too much credit sometimes, thanks, that fix makes a lot of sense.

Issue #328, opened 2015-12-17
Using AirBnB's Enzyme - help/question


SpencerCDixon:
Was just wondering if anyone has been able to get Enzyme working with this starter kit or if there are plans to use it in the future?  When I tried, karma was throwing all sorts of errors and I couldn't get it working.  My knowledge of karma and testing in JS in general is pretty limited though so maybe there's an obvious fix I dont know about.

I get a whole bunch of these errors:
```
ERROR in ./~/entities/maps/xml.json
Module parse failed: /Users/spencerdixon/Dropbox/CurrentProjects/React/enzymeTesting/node_modules/entities/maps/xml.json Line 1: Unexpected token :
You may need an appropriate loader to handle this file type.
| {"amp":"&","apos":"'","gt":">","lt":"<","quot":"\""}
|
 @ ./~/entities/lib/decode.js 3:16-43

ERROR in ./~/htmlparser2/~/entities/maps/decode.json
Module parse failed: /Users/spencerdixon/Dropbox/CurrentProjects/React/enzymeTesting/node_modules/htmlparser2/node_modules/entities/maps/decode.json Line 1: Unexpected token :
You may need an appropriate loader to handle this file type.
| {"0":65533,"128":8364,"130":8218,"131":402,"132":8222,"133":8230,"134":8224,"135":8225,"136":710,"137":8240,"138":352,"139":8249,"140":338,"142":381,"145":8216,"146":8217,"147":8220,"148":8221,"149":8226,"150":8211,"151":8212,"152":732,"153":8482,"154":353,"155":8250,"156":339,"158":382,"159":376}
 @ ./~/htmlparser2/~/entities/lib/decode_codepoint.js 1:16-46

ERROR in ./~/entities/maps/decode.json
Module parse failed: /Users/spencerdixon/Dropbox/CurrentProjects/React/enzymeTesting/node_modules/entities/maps/decode.json Line 1: Unexpected token :
You may need an appropriate loader to handle this file type.
| {"0":65533,"128":8364,"130":8218,"131":402,"132":8222,"133":8230,"134":8224,"135":8225,"136":710,"137":8240,"138":352,"139":8249,"140":338,"142":381,"145":8216,"146":8217,"147":8220,"148":8221,"149":8226,"150":8211,"151":8212,"152":732,"153":8482,"154":353,"155":8250,"156":339,"158":382,"159":376}
 @ ./~/entities/lib/decode_codepoint.js 1:16-46
```

[...]

PR #329, opened 2015-12-17
feat(json-loader): add dependency and configure webpack

davezuko:


davezuko:
@hairychris it comes from this: https://github.com/davezuko/react-redux-starter-kit/issues/328

It's a pretty trivial dependency to add, so I figured it would make some people's lives easier.

hairychris:
Sounds good to me, thanks for clearing that up.

Issue #321, opened 2015-12-16
Should We Deprecate `npm run dev:nw`?


davezuko:
As pointed out in https://github.com/davezuko/react-redux-starter-kit/issues/317 by @sars, would this be a good move for the starter kit? I've personally never use the `dev:nw` command, and am not sure how many people actually do. If it's not a critical feature, I'd suggest we consider eliminating it from the core of the repo and point users to the extension if they want a similar experience.

Note: The sidebar dock would still remain.

mtsr:
I do use it, since I often work on just the 13” screen at home, but the extension might be a good alternative, although I haven’t tried.

davezuko:
Ah ok, thanks @mtsr. You're a pretty core contributor so I don't want to remove this if you've found it useful. That said, I'd be curious to see what you think of the extension; I'll give it a try myself.

sarunast:
I am using it as well. But the main reason for me that dev tools always appears on screen if I refresh the window, if by default somehow it would be possible to hide and open it with ctrl+h, then I wouldn't need the dev window. Basically my dev window is clean but whenever I am interested in states I peak in the other window, I don't want to see them all the time or always hide it with ctrl+h.

davezuko:
@sarunast there is a way to do exactly that: https://github.com/gaearon/redux-devtools/issues/44#issuecomment-137578200

sarunast:
@davezuko thanks you very much. I was thinking that there should be an easy way.

neverfox:
I *only* use `dev:nw` lol (ok, not always). When doing mobile-first development, it's annoying to have to hide the devtools on each hard refresh and it's unreadable, in any case, on small viewports.

neverfox:
Is your main concern the extra code needed to support it?

davezuko:
@neverfox that relates to an old issue I posted where I wanted to set the default visibility of the devtools to hidden, but was outvoted. Maybe this would be a good chance to make it a configuration option.

And yeah, it's just that it's yet more code in the starter kit. I guess it's not a huge deal, and from what I'm hearing people do use it, so I'm open to keeping it in the project.

mtsr:
With at least two solutions (extension and default hiding + hotkey) don’t feel the need to keep it for me. I still like having it, though I think the extension will replace it nicely for me.

patrickheeney:
[...]

PR #337, opened 2015-12-21
Added in hooks for chrome's redux dev tools

CurtisHumphrey:
see https://github.com/zalmoxisus/redux-devtools-extension for more details

### DIFF
--- a/src/containers/Root.js
+++ b/src/containers/Root.js
@@ -18,7 +18,9 @@ export default class Root extends React.Component {
   }
 
   get devTools () {
-    if (__DEBUG__) {
+    if (window.devToolsExtension) {
+      return null; // will be loaded in chrome's redux dev tools
+    } else if (__DEBUG__) {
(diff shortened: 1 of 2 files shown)


patrickheeney:
Related #321. Maybe the extension could be loaded in by default (e.g. removing the old dev tools) because not a lot of people expressed interest in the devtools in the viewport option.

davezuko:
@patrickheeney I use the sidebar tools a lot and am a fan of having them be the default just so that people who spin the app up for the first time see that devtools do in fact work.

@CurtisHumphrey looks good, thank you. There's a lint error that's failing the CI build right now, so if you could fix that then we can get this merged.

davezuko:
Closing in favor of https://github.com/davezuko/react-redux-starter-kit/pull/356, but thanks for bringing this up!

CurtisHumphrey:
You are welcome.

PR #345, opened 2015-12-23
High CPU usage Sublime

lukasz-madon:
While compiling in "development" settings Sublime has a hard time parsing app.js file

```
worker 85850 appears stuck while processing file react-redux-starter-kit/dist/app.899b349b9005e200c0f4.js, killing process
```

as a solution exclude `dist` folder

### DIFF
--- a/README.md
+++ b/README.md
@@ -254,3 +254,10 @@ Troubleshooting
 This is most likely because the new window has been blocked by your popup blocker, so make sure it's disabled before trying again.
 
 Reference: [issue 110](https://github.com/davezuko/react-redux-starter-kit/issues/110)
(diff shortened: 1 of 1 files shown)


davezuko:
What file does that go in? An .editorconfig or something else?

lukasz-madon:
Sadly, .editorconfig is not able to support this. It goes into `Library/Application Support/Sublime Text 3/Packages/User/Preferences.sublime-settings`

davezuko:
I might actually be moving this into a better FAQ that is a WIP, so holding off on the PR for now. Definitely something worth nothing, though.

PR #352, opened 2015-12-24
add proxy support

asafamr:
added an option to proxy requests to another address by prefix. e.g. send /api/* requests to a dev. backend running at a different port.

relatively new to webpack so I hope I did it right


patrickheeney:
Couple of questions:

Where is the api server running at port 1337, is this a default port for http-proxy?

Do you have to wrap every call around filterProxy? I thought the routes were executed in order of being defined so the later routes shouldn't hit the proxy?

asafamr:
1337 is the default port for sailsJS and I use it for the back end server.

the express.use function registers middleware for all routes so this way we can disable it for proxy requests. inspired by 
http://stackoverflow.com/questions/19337446/is-it-possible-to-disable-remove-a-middleware-for-specific-route-in-expressjs

davezuko:
Sorry, it's been crazy with the holidays. Thanks for the PR, I haven't had a chance to really look at it yet but will once things calm down.

chimurai:
Consider using [`http-proxy-middleware`](https://github.com/chimurai/http-proxy-middleware).

It's build on top of the same `http-proxy`, but with some sugar syntax for easy middleware configuration, path/pattern matching/exclusion and much more.

asafamr:
nice. changed to http-proxy-middleware and merged new commits
@davezuko you guys are super responsive good job

davezuko:
Just as an FYI been holding off on this since I'm planning to migrate back to Koa, just working some things out. https://github.com/davezuko/react-redux-starter-kit/pull/379, probably will be a few days before a merge.

davezuko:
Hey sorry, I've been incredibly swamped with work this past week; normally it doesn't take me this long to get to PR's. Koa has replaced Express in the kit, but if you think it would be worthwhile to resubmit this to work w/ the new framework I'd be happy to discuss it. Really appreciate the contribution, sorry about the last minute change.

kitze:
@asafamr have you found a solution to this with koa?

yiziz:
@kitze If you're still looking for a solution, take a look at https://github.com/davezuko/react-redux-starter-kit/pull/478

PR #355, opened 2015-12-26
chore(webpack): use ProvidePlugin to reduce boilerplate

terakilobyte:
I really appreciate the effort that has gone into this project! It's been my goto to test new ideas and build a few projects.

Submitting this PR for review to reduce boilerplate. It uses webpack's ProvidePlugin to implicitly import React and ReactDom.

It also turns the `react-in-jsx-scope` rule off and makes React and ReactDOM globals.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -11,9 +11,12 @@
     "__DEV__"      : false,
     "__PROD__"     : false,
     "__DEBUG__"    : false,
-    "__DEBUG_NEW_WINDOW__" : false
+    "__DEBUG_NEW_WINDOW__" : false,
+    "React": true,
+    "ReactDOM": true
   },
   "rules": {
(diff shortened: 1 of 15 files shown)


davezuko:
Very cool, I like it a lot. My one concern with this approach is that it potentially makes it more difficult for people to create isomorphic apps without pre-bundling it with webpack, but I guess CSS modules currently present a similar problem so I can't really complain. I'll take a look at this tomorrow and probably get it merged in :+1:.

One suggestion, what about putting this in config? Something like:

```
compiler_globals: [...]
```

terakilobyte:
Makes more sense than where I initially put the configuration. I've updated this PR with the changes.

davezuko:
:+1: merged, thank you.

PR #358, opened 2015-12-27
fix(cssnano): Don't use unsafe cssnano options

SimonSelg:
I was including a local (woff-)version of the Roboto font using `@font-face` in core.scss, because [react-toolbox](https://github.com/react-toolbox/react-toolbox) requires it to be available. However, it would not be available, nor would the `@font-face` show up in the css file that is generated after running compile in an production environment.

Turns out, it's because of cssnano. It has some [*unsafe* options](http://cssnano.co/optimisations/), for example [postcss-discard-unused](https://github.com/ben-eb/postcss-discard-unused), which removes `@counter-style`, `@keyframes` and `@font-face`  if it does not find any usage of the declared keyframe/font/counter-style in the same file.

Since we're using css-modules, the usage of a declared font or animation is very likely *not* to be in the same file, for example in my case Roboto was being used in some components scss files. But postcss-discard-unused removed the `@font-face` entry in this case, since there is no usage of Roboto in core.scss.

To solve this problem, we should disable the unsafe options.

[postcss-zindex](https://github.com/ben-eb/postcss-zindex), [postcss-merge-idents](https://github.com/ben-eb/postcss-merge-idents) and [postcss-reduce-idents](https://github.com/ben-eb/postcss-reduce-idents) are unsafe option, and they do not work well with multiple css files.

By setting `save`to `true`, we disable all 4 unsafe options ([source](https://github.com/ben-eb/cssnano/issues/28)).

For more information check http://cssnano.co/optimisations/

### DIFF
--- a/build/webpack-environments/_base.js
+++ b/build/webpack-environments/_base.js
@@ -125,6 +125,7 @@ const webpackConfig = {
         remove: true,
         browsers: ['last 2 versions']
       },
+      safe: true,
       discardComments: {
         removeAll: true
       }


davezuko:
Awesome, never would have known about that. Just curious, do you have an opinion on using CSS modules vs. your typical sass/etc. configuration? I've recently started questioning using CSS modules by default due to a few issues:

1) It doesn't play well with HMR; you have to explicitly write the HMR code yourself for replacing imported classes.
2) It doesn't play well with Isomorphic rendering, since it requires CSS to be preprocessed with webpack and the classnames are dynamic, meaning you can't just ignore all `.(css|scss)` files with `require.extensions`.

I know this isn't exactly relevant to the PR, but if you have similar feelings I'd consider making a general poll to see how others feel.

SimonSelg:
To say this first: You probably have far more experience with react and the whole ecosystem then I have. And since we're talking about a React.js specific usage here (react is based on components, which influences the way I'd like to behave my CSS, see the latest paragraph) I think this might matter.

About isomorphic rendering: You are right, it does not play well with universal/isomorphic rendering *without any configuration*. But there is [webpack-isomorphic-tools](https://github.com/halt-hammerzeit/webpack-isomorphic-tools), which kind of solves this problem (it's used by the [other big react.js boilerplate](https://github.com/erikras/react-redux-universal-hot-example), which has universal rendering but heavily dictates a structure). It's still not perfect, and I'd love to have something that *just works out of the box*, but I don't see a good alternative there. 

I have not yet researched the HMR problematic well, so I don't really have an opinion on that one. Will do some research. 

I like this approach a lot better tho then using inline styles, to [quote the creator of react-toolbox](https://www.reddit.com/r/reactjs/comments/3rfaqv/react_toolbox_bootstrap_your_application_with/cwnwm3d) here:

> Inline styling: Among other stuff, when you use inline styles you are missing a lot of great features like pseudoelements, pseudoselectors, media queries, targeting tags, caching... etc etc.

It just feels wrong to me. And when used in combination with react components, customization of components (for example components from [materialize-ui](https://github.com/callemall/material-ui), which uses inline styles) is hard, since inline styles have a higher priority then non-inline styles, you have to use `!important` everywhere (or use inline styles for that purpose too ofc... but then integration on other project that don't use inline styles, and that's still the majority, is difficult).

Of course one could just use a normal sass setup here, but for me there is a key problem: The scope. When writing components, I'd like them to behave the same in every environment, and I don't want to worry about the classname 'super-fancy-button' being taken. Except from using webcompontents/polymer, which does not work (as far as I know) together with react, I don't know any other way to solve this problem.

So from my POV, the HMR downside is worth the benefits over the other solutions that I know of. 

Of course, if there are any better ways I'd love to hear about them and to incorporate them into my stack.

davezuko:
Merging this, thanks for the detailed response, I'll reply when I get a chance to write something meaningful, haha.

SimonSelg:
Regarding the isomorphic/universal rendering issue: There is now [https://github.com/istarkov/babel-plugin-webpack-loaders](this). Looks awesome to me!

PR #359, opened 2015-12-27
Extend section about global variables with info about automatic imports

sbusch:
... and add subtitle for ProvidePlugin within section "Provided Plugins"


davezuko:
Thank you, was just thinking about doing this myself!

PR #367, opened 2015-12-28
Relative assets fix (fixes #357)

### LINKED ISSUE #357: Assets referenced with url don't work when compiled in a development environment
Hey,

please excuse the confusing title, I couldn't think of a better one.

In webpack-style-loader there is an issue with assets that are referenced using `url(...`) (for example images or fonts) when style loader is used with ?sourceMap option ([explanation of this issue on Stack Overflow](http://stackoverflow.com/questions/34133808/webpack-ots-parsing-error-loading-fonts/34133809#34133809), [note about this on the style-loader repo](https://github.com/webpack/style-loader/blame/master/README.md#L73) ).

This is [correctly solved](https://github.com/davezuko/react-redux-starter-kit/blob/f0176c3e6e9bbffc3a2a28e44c2274258581070a/config/_development.js#L10-L16) by this boilerplate when Hot Module Replacement is enabled, meaning when the build process is started using `npm run start`, `npm run dev:nw` or `npm run dev:no-debug`.

But when using `npm run compile` with `NODE_ENV` *not* set to `production`, Hot Module Replacement will be disabled (=> the [fix does not apply anymore](https://github.com/davezuko/react-redux-starter-kit/blob/f0176c3e6e9bbffc3a2a28e44c2274258581070a/config/_development.js#L10-L16)), but source-maps are still enabled. This triggers the issue meaning that all assets that are referenced using `url(...)` don't load correctly anymore. 

I've been thinking about how to solve this issue, and I'm not sure how to do it. We could 
* either use extract-text-webpack-plugin when running `npm run compile` in a *development* environment (this solves the issue, as explained in the [Stack Overflow post](http://stackoverflow.com/questions/34133808/webpack-ots-parsing-error-loading-fonts/34133809#34133809)), currently it is only unsed in a *production* environment
* or add an entry to the config where the user can specify the absolute path for compile builds

The last option would be 'unportable' tho, because the compiled bundle will only work on this particular absolute path, which is not intuitive from my POV.

Either way, this should be solved or at least documented in the README, because this issue might be hard to debug for people who stumble on it. I can create a PR for it if we decide on a solution.

SimonSelg:
This fixes issue #357

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ Features
   * sass-loader
   * postcss-loader with cssnano for style autoprefixing and minification
   * Bundle splitting for app and vendor dependencies
-  * CSS extraction during production builds
+  * CSS extraction during builts that are not using HMR (like `npm run compile`)
   * Loaders for fonts and images
 * [Express](https://github.com/strongloop/express)
   * webpack-dev-middleware
--- a/build/webpack-environments/_base.js
+++ b/build/webpack-environments/_base.js
@@ -1,6 +1,7 @@
 import webpack from 'webpack'
 import cssnano from 'cssnano'
 import HtmlWebpackPlugin from 'html-webpack-plugin'
+import ExtractTextPlugin from 'extract-text-webpack-plugin'
 import config from '../../config'
 import _debug from 'debug'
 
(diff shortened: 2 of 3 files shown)


davezuko:
:+1: much appreciated, I'll pull it down and test locally later tonight and then get this merged.

SimonSelg:
Oops.  Must have overlooked [this](https://github.com/davezuko/react-redux-starter-kit/commit/ea19eed218ca6c2e14d2e85f2fa33837c7ae82a2). 

Any idea why travis (=> eslint) didn't complain about that semicolon?

davezuko:
@SimonSelg Travis is only linting the app source code atm, I'll have to add a task to lint the app as well.

PR #370, opened 2015-12-30
Make publicPath truly relative and configure history for basename

neverfox:
A truly relative publicPath and the ability to configure a base path for React Router helps when deploying app in environments where its not served from the root path of the web server. The combination of changes lets you put the app arbitrarily deep, while still successfully serving resources and keeping route definitions defined in terms of '/'.


davezuko:
Awesome, thank you!

PR #371, opened 2015-12-30
Modified compiler_public_path to undefined

yosqueoy:
I suggest to change the default value of compiler_public_path from '/' to undefined. Current value doesn't work when proxy sets prefix.

Current version creates the line below. The src is an absolute path, so it will not work if prefix like /deploy is set by proxy. 
```
<script src="/app.c3e1585f9dcb449f54f3.js"></script>
```
Proxy expects http://<host>/deploy/app.c3e1585f9dcb449f54f3.js, but the src is  http://<host>/app.c3e1585f9dcb449f54f3.js.

After changing compiler_public_path to undefined, the line becomes a relative path, which works well with proxy prefix. 
```
<script src="app.c3e1585f9dcb449f54f3.js"></script>
```

### DIFF
--- a/config/_base.js
+++ b/config/_base.js
@@ -35,7 +35,7 @@ const config = {
   compiler_hash_type       : 'hash',
   compiler_fail_on_warning : false,
   compiler_quiet           : false,
-  compiler_public_path     : '/',
+  compiler_public_path     : undefined,
   compiler_stats           : {
     chunks : false,
     chunkModules : false,


davezuko:
Hey there, just woke up; is this related to https://github.com/davezuko/react-redux-starter-kit/pull/371? I still have to re-read both PR's but it seems so.

davezuko:
https://github.com/davezuko/react-redux-starter-kit/pull/370 has been merged; let me know if it doesn't solve the same problem.

neverfox:
Yes, #370 was designed to fix just that. I didn't consider undefined as a possible initial value as I didn't know if it might lead to some bad side effects. Is there some advantage to that over empty string?

yosqueoy:
Thanks, #370 fixed my issue. Empty string is perfectly fine and that has the same effect as undefined because webpack internally uses publicPath option as below.
```
var publicPath = outputOptions.publicPath || "";
```
https://github.com/webpack/webpack/blob/9f440e30ecac00bfc27b91d372a969d3414d194c/lib/TemplatedPathPlugin.js

PR #377, opened 2016-01-01
Add localization using react-intl

juanda99:


davezuko:
Closing this for now, but will add a README section to point to your branch as an example if you plan on keeping it live.

PR #379, opened 2016-01-02
feat(koa): replace express with koa

davezuko:


neverfox:
Koa now seems to encourage using `async/await` instead of `yield` when using ES7.

SimonSelg:
@neverfox : Well, this applies to Koa 2.0 only from what I know. [That's also why they are not releasing it yet](https://github.com/koajs/koa/issues/631), because they are heavily using async/await and are waiting for the specs to be finalized (async/await is only in the proposal stage yet) and for v8 (and therefore node) to implement it. In Koa 1.0 you explicitly have to enable the use of async/await using the experimental flag.

Besides from that, I worked with async/await yesterday, and it's super cool from my pov and helps a lot to cut down complexity in your code. In order to use async/await with node, you need to use babel runtime transpiling or use babel-node.

Also, it's only in the *proposal* stage meaning it *could* not make it into the final release of es7. 

But it simplifies asynchronous code greatly without sacrificing performance (even in the transpiled babel code), so it sounds like an awesome feature to me.

davezuko:
@neverfox Yeap, definitely want to be using Koa 2, but most of the middleware hasn't caught up yet so figured we can stay with stable 1.0 for now.

neverfox:
Yeah, I probably would avoid it for those reasons for now. Besides, it's a deep rabbit hole to go down for a simple server.

pkieltyka:
Cool. Just wondering though, what is the benefit of using Koa instead of Express for this kit? the server is meant just to serve very basic contents and stays pretty much the same through the life of a project.

davezuko:
@pkieltyka I find Koa a lot easier to use when writing server functionality -- either API's or isomorphic rendering -- so while it really isn't a meaningful switch out of the box for the starter kit, I think it improves the experience for adding new features.

PR #381, opened 2016-01-02
style(standard): DRY out PropTypes

neverfox:


davezuko:
If we're going to go this route I'd rather destructure on the import, eliminates even more code.

`import { Component, PropTypes } from 'react'`.

I understand that your method was done since the `Provide` plugin is being used to expose React globally, but I'd consider dropping that plugin to eliminate this oddity. Personally I didn't really consider the inclusion of the ProvidePlugin as thoughtfully as I should have, and I'd really like to steer the project away from too much webpack magic. If you'd like to wait for me to drop it before making this change that might be best, unless you have other thoughts.

neverfox:
You're right, I did it in light of the ProvidePlugin because normally I do:

```
import React, { PropTypes } from 'react'
```

I'm fine with resubmitting something in that vein. It's a useful demonstration too, since some people just getting into ES6 don't know you can do that.

neverfox:
Also, I haven't checked, but does this global injection carry over to tests? If not, that's a good reason to drop it.

davezuko:
It does carry over into tests, so no problems there. I'll mull it over a bit tonight and figure out whether or not to drop the ProvidePlugin, and we can go from there.

patrickheeney:
+1 for `import React, { PropTypes } from 'react'`. `ProvidePlugin` is a lot of magic and in some IDEs throws errors because of missing imports.

davezuko:
I will get to this tomorrow :)

SimonSelg:
+1  for that from me too. Also in IDEs that support auto completion (like Webstorm, specificly when you add typescript definitions for react) it breaks the autocompletion for React and PropTypes.

davezuko:
By "tomorrow" I mean hopefully by the end of this week. ;)

davezuko:
Reverted to pre-ProvidePlugin w/ a few additional tweaks: https://github.com/davezuko/react-redux-starter-kit/pull/410

neverfox:
Updated PR

davezuko:
Thank you, appreciate it.

PR #387, opened 2016-01-03
fix: remove node_modules from the coverage report

SpencerCDixon:
For some reason an npm package was getting included in my coverage reports.  Adding this `exclude` made it stop running coverage on the npm package.  The package was `js-cookie` in case you were wondering.  Not sure if this is actually a necessary PR or if I was just doing something silly in my code.


SpencerCDixon:
Also just realized there is a contributions guide.  My bad for totally messing this up.  Can close and resubmit if you'd like/think this is worthwhile

davezuko:
@SpencerCDixon thanks for the PR, and don't worry about it, I'll just rebase and alter the commit message.

PR #396, opened 2016-01-04
refactor(webpack): consolidate configuration

davezuko:
/cc @patrickheeney 

This is the first step to consolidating webpack configuration. A future PR will clean this up even further once `better-npm-run` allows us to always run Karma w/ `NODE_ENV === test`. There will be a lot of volatility with the webpack system as we move toward something more configurable (per our discussions), but this is just the first stage of that.


davezuko:
Merged.

PR #402, opened 2016-01-07
style(lint): whitespace and semicolon fixes

AndreiRailean:


src/containers/Root.js
```diff
@@ -6,7 +6,7 @@ export default class Root extends React.Component {
     history: React.PropTypes.object.isRequired,
     routes: React.PropTypes.element.isRequired,
     store: React.PropTypes.object.isRequired
-  }
+  };
```
justingreenberg: @davezuko this resolves an issue i was about to create—[as of yesterday](https://github.com/babel/babel/commit/f31099f383b52cf4fe1786188f6421529dea865b), babylon's parser will throw if a ClassProperty node [does not end in a semicolon token](https://github.com/babel/babel/blob/master/packages/babylon/src/parser/statement.js#L751)... which is gonna be rough for standard js until there's native class support. you could open an issue on babel's phabricator and make a case for removing ClassProperty semi validation (since it's still stage 1 experimental) but they probably have a good reason for enforcing it

davezuko: Yeah I can't believe they released this as a minor version without any real warning. Was triaging this all day at work, fun times :(

davezuko:
Ty @AndreiRailean, had to rebase this and it broke git's auto merge, but this is now in master.

PR #411, opened 2016-01-08
feat(NotFoundView): replace AboutView with 404 redirect demo

davezuko:


justingreenberg:
fwiw, i think it's a little confusing having an actual `/404` route... and this doesn't really demonstrate redirect functionality. i think it might make more sense to leave the AboutView with a link from homepage, and then include another link "Go to BLAHBLAHBLAH" and *actually* link to an unknown route to demonstrate redirect functionality... thoughts?

justingreenberg:
the more i think about it, this isn't how Redirect is meant to be used... from [the docs]
(https://github.com/rackt/react-router/blob/master/docs/API.md#redirect)....

>A \<Redirect\> sets up a redirect to another route in your application to maintain old URLs.

the correct way to catch 404s is to map a path to wildcard in root:

```js
  <Route path='/' component={CoreLayout}>
    <IndexRoute component={HomeView} />
    <Route path='/about' component={AboutView} />
    <Route path='*' component={NotFoundView} />
  </Route>
```

ie. this way `<Link to='/fakeroute'>Broken Link</Link>` will render `NotFoundView` component, but the url will still represent the intended target `https://example.com/fakeroute`, which is the expected bahavior from user's perspective

i'm sitting down for dinner now, i'll put together a pr when i'm back at my desk

davezuko:
That route structure makes sense, but only if you don't want an actual /404 landing page -- which is totally up to the implementor. I had the functionality of some websites in mind where they _do_ in fact have a `/404` landing page (www.bloomberg.com/404) but I'm willing to reconsider.

The link to the /404 page is strange, I admit, but I don't think another placeholder view achieves much.

Issue #408, opened 2016-01-08
Question? Can I disable css-loader's cssmodules rewriting rules for @import statements?


pixeldrew:
For example, I have a css library installed via npm and would like to include it into my core styles but it looks like the webpack module css-loader is rewriting all the rules.

I want to keep the cssmodules rewriting for other css files included from the view layer but for the core styles  i'd like to have the rules not renamed.

I don't see any settings in css-loader. Does anyone know a webpack module option trick to do this?

xquezme:
In your:
```
./src/styles/core.scss
```
Try this
```css
:global {
  @import './../node_modules/YOUR_LIB';
}
```

pixeldrew:
Hi Sergey,

I did try that, the library imports but the rules are still rewritten.

```scss
:global {
    @import 'npm-module/css/aframe-work.css';
}
```

bezidejni:
Doesn't seem to work for me either, I'm using v1.0.0

neverfox:
I just change my webpack config to treat css files outside of `src` as global CSS. This should probably be the default way the kit is set up as there are several scenarios where third-party libraries require you to use global CSS even though you use modules for your own components. I'm not a fan of the current either/or setup. I'll probably make a PR for this.

davezuko:
I believe @neverfox 's latest PR (which has been merged) fixes this. Please let me know if it doesn't and I will reopen.

pixeldrew:
This is still happening in master.

pixeldrew:
I can't include a module into core.scss, am I supposed to include it somewhere else?

pixeldrew:
Doh, ignore me. Seems the module I was trying to load was already rewritten...

pixeldrew:
Apologies for the back and forth on this ticket, yes this is still happening.

neverfox:
@pixeldrew Just so I understand, you're just having a problem importing a third-party CSS library into core?

pixeldrew:
even not wrapping it in global this still doesn't work:

I forked master (added an otf font fix on my branch):
```sh
git clone https://github.com/pixeldrew/react-redux-starter-kit
cd react-redux-starter-kit/
npm install
npm install --save topcoat
echo "@import 'topcoat/css/topcoat-desktop-light.css';" >> src/styles/core.scss 
export NODE_ENV='production' && npm run compile
```

The output is as follows (note that the topcoat file goes through the css-loader with localIdentName=...)

```sh
[...]

PR #417, opened 2016-01-09
refactor(webpack): allow mix of global and module CSS

neverfox:
Addresses things like #408


geohuz:
I'm using 0.18.0, is there anyway to apply your change?

davezuko:
@geohuz you should just be able to put the additional loaders in https://github.com/davezuko/react-redux-starter-kit/blob/v0.18.0/build/webpack/_base.js and then add the `include` property to the existing style loaders.

PR #418, opened 2016-01-10
fix(chore): devTools - add ReactDOM, React to devTools

jaeh:
npm run dev:nw failed before this fix


davezuko:
Thank you!

jaeh:
thanks for doing all the prior work that was needed to enable me to add this tiny fix :)

PR #419, opened 2016-01-10
added in settings so jsx files could be used too

CurtisHumphrey:


davezuko:
Made some pretty major changes here: https://github.com/davezuko/react-redux-starter-kit/pull/436 and included your change in it since it created a merge  conflict.

This also changes how the test suite works, if you want to see whether or not it fixes some of your coverage issues.

PR #420, opened 2016-01-10
fix: grammar in CoreLayout

simplyianm:


davezuko:
Thank you, been meaning to fix that for a while now.

PR #442, opened 2016-01-13
feat(webpack): component name resolver

justingreenberg:
simple plugin to resolve nested components with filenames that match directory names... instead of:
```js
// full path
import CoreLayout from 'src/layouts/CoreLayout/CoreLayout.js'
```
```js
// using root and extension resolvers (current setup)
import CoreLayout from 'layouts/CoreLayout/CoreLayout'
```
after this PR, we can use the shortcut:
```js
// using component resolver plugin
import CoreLayout from 'layouts/CoreLayout'
```
standard import syntax and behavior are not affected, so the former methods will still work

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,18 +180,43 @@ You can redefine which packages to bundle separately by modifying `compiler_vend
 ]
 ```
 
-### Webpack Root Resolve
-Webpack is configured to make use of [resolve.root](http://webpack.github.io/docs/configuration.html#resolve-root), which lets you import local packages as if you were traversing from the root of your `~/src` directory. Here's an example:
+### Webpack Resolvers
+Webpack is configured to make use of several shortcuts that help make imports more flexible, modular and DRY:
+ - [resolve.root](http://webpack.github.io/docs/configuration.html#resolve-root) lets you import local packages as if you were traversing from the root of your `~/src` directory.
+ - [component-resolver-webpack](https://github.com/toptal/component-resolver-webpack) shortcut for resolving components with filenames that match their directory names, the widely accepted convention for modular react components used in this projecs.
 
+**Using `resolve.root`:**
 ```js
-// current file: ~/src/views/some/nested/View.js
-
 // What used to be this:
-import SomeComponent from '../../../components/SomeComponent'
+import SomeComponent from '../../../components/SomeComponent/SomeComponent.js'
 
 // Can now be this:
-import SomeComponent from 'components/SomeComponent' // Hooray!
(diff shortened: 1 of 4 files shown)


thangngoc89:
Please don't accept this PR. This will make testing with mocha in node very hard.

justingreenberg:
@thangngoc89 can you explain please?

thangngoc89:
I usually test redux modules (and other stuff) with mocha run in node environment (no karma). I can't find a way to implement webpack with mocha to resolve path. But with current root resolver setup in webpack, I can use this solution http://stackoverflow.com/a/31334506 .

If you want to achieve the same result like your PR. You can create a `layouts/CoreLayout/index.js` with the following content:

```js
import CoreLayout from './CoreLayout'
export default CoreLayout
```

justingreenberg:
i mean, that's because you are *choosing* to bypass webpack and use mocha directly... does this PR affect anything internal to the starter kit? i'm sorry if this PR affects your personal workflow, but the project goal is to provide a starter with the best developer ergonomics out of the box. 

after all, the kit is **built on webpack** and **configured for karma**.
 
using mocha directly was definitely popularized by the [redux docs](http://redux.js.org/docs/recipes/WritingTests.html). personally, i [like the simplicity of tape](https://medium.com/javascript-scene/why-i-use-tape-instead-of-mocha-so-should-you-6aa105d8eaf4) (with enzyme) and would love to see our testing move in that direction, but that's up to @davezuko 

also, with respect to your testing issue:
 - the [solution you are using](http://stackoverflow.com/a/31334506) is pretty hacky—it's monkey patching NODE_PATH at runtime to trick node's internal module resolver, which is unnecessary when you have the full power of webpack at your fingertips. is there a reason you are opposed to using karma?
 - if you still wanted to use that solution with this PR, there's no need to add index.js to every component folder—just remove the resolver plugin and use full import paths

thangngoc89:
@justingreenberg : thank you for your reply 

> which is totally unnecessary when you have the full power of webpack at your fingertip

I can't find a good tutorial for integrating mocha with webpack in node environment

>  is there a reason you are opposed to using karma

no. I am not opposed to using Karma. I use mocha directly in my laptop and karma in CI. Karma is just too slow for this.

And thank you for your great tips with ES6 module import/export

davezuko:
Closing this since, as nice as it is (I really do despise the duplicate names), I think we should limit webpack magic where it makes sense. `root.resolve` is arguably similar, but I find it slightly more intuitive than this.

@justingreenberg when we get 2.0.0 out I will 100% link to this PR though as a webpack recipe for cleaning up some code smells for users who want it.

PR #448, opened 2016-01-14
fix(webpack): re-add loader for css

vkvelho:
For some reason this loader was dropped on consolidating configuration commit:

https://github.com/davezuko/react-redux-starter-kit/commit/a86512777294eebba1349fc5a2d692c8e71726c1#diff-2aa085fcb2f0f27da800b7ffc26fb9d9L91

### DIFF
--- a/build/webpack.config.js
+++ b/build/webpack.config.js
@@ -141,6 +141,16 @@ webpackConfig.module.loaders.push({
   ]
 })
 
+webpackConfig.module.loaders.push({
+  test: /\.css$/,
+  include: /src/,
+  loaders: [
+    'style',
+    cssLoader,
+    'postcss'
+  ]
+})
+
 // Don't treat global SCSS as modules
 webpackConfig.module.loaders.push({
   test: /\.scss$/,


davezuko:
Thanks for the re-add, but would you mind submitting it with just the loader? I don't think we need an additional sample CSS file.

vkvelho:
Sure, but that CSS (not SCSS) file would've caused an error on testing and prevailed the missing loader, so I thought it was good idea to prevent the same issue happening again in the future.

Another way to test the same case could be changing the HomeView.scss to HomeView.css, because that file content is plain css anyway. Maybe add a comment that the file could be also scss.

I'll fix the PR after your opinion on that.

davezuko:
@vkvelho I appreciate the foresight, but I think that issue should be solved through better tests against the kit itself rather than making the application itself a unit test, if you know what I mean. I'm guilty of this myself, so no worries.

I do like your other recommendation, however, though I'd suggest noting that the `.scss` file could just be a regular `.css` file, simply because that way people can immediately see more features out of the box.

vkvelho:
@davezuko Yeah, I hear you. Time is very limited resource, so plain loader commit as wished it is.

davezuko:
Had to rebase this locally, so the auto merge detection broke. That said, it's been merged, thank you for your help!

vkvelho:
@davezuko I force pushed the commit to keep the history clean. You can ignore this if that's cool, but if it was the reason for rebase or extra work and it was better if I did something different, please let me know.

davezuko:
@vkvelho I actually appreciate that you force pushed; I had to do a force push myself after a wonky merge where I somehow screwed up rewording some other commits, so it was my fault. Either way, it's in master now.

PR #457, opened 2016-01-15
fix(karma): add sourcemap to get better references

vkvelho:
All the errors and failed tests of Karma reference just to the `test-bundler.js` and I think that's so useless information for a developer that I consider it a bug. For example:

```
(View) Home
    ✓ Should render as a <div>.
    ✗ Should include an <h1> with welcome text.
	Error: Did not find exactly one match for tag:h1
	    at /Library/xxxxx/react/tests/test-bundler.js:48953
	    at /Library/xxxxx/react/tests/test-bundler.js:31020
```

I was able to fix that using `karma-sourcemap-loader` and now the output looks like this:

```
(View) Home
  ✓ Should render as a <div>.
  ✗ Should include an <h1> with welcome text.
	Error: Did not find exactly one match for tag:h1
	    at /Library/xxxxx/react/tests/test-bundler.js:48953 <- webpack:///~/react/lib/ReactTestUtils.js:206:
	    at /Library/xxxxx/react/tests/test-bundler.js:31020 <- webpack:///tests/views/HomeView.spec.js:37:4
```

I really don't know how far from perfect the fix is, but the issue is gone!

### DIFF
--- a/build/karma.conf.js
+++ b/build/karma.conf.js
@@ -19,7 +19,7 @@ const karmaConfig = {
   singleRun: !argv.watch,
   frameworks: ['mocha', 'chai-sinon', 'chai-as-promised', 'chai'],
   preprocessors: {
-    [`${config.dir_test}/test-bundler.js`]: ['webpack']
+    [`${config.dir_test}/test-bundler.js`]: ['webpack', 'sourcemap']
   },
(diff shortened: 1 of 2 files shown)


davezuko:
Ah, yes! You are awesome, thank you!

PR #458, opened 2016-01-15
feat(router): migrate routing ecosystem (v2)

justingreenberg:
combines #452 and #453, since the library APIs are in sync as of v2. updates `react-router`, `history`, and `redux-simple-router` to new APIs and wires everything together.

**the instantiated history object is the point of contact between router and store**

 - sync with `react-router` in main.js (useBasename is baked in to useRouterHistory HOC)
 - pass instantiated history to configureStore.js, subscribe `redux-simple-router` middleware
 - configure middleware to replay actions in \_\_DEBUG\_\_ mode (for devtools support)

let me know if you have any questions :)


src/redux/configureStore.js
```diff
@@ -1,29 +1,27 @@
...
-  createStore
-} from 'redux'
 
-export default function configureStore (initialState) {
-  let createStoreWithMiddleware
-  const middleware = applyMiddleware(thunk)
+function withDevTools (middleware) {
+  const devTools = window.devToolsExtension
+    ? window.devToolsExtension()
+    : require('containers/DevTools').default.instrument()
```
maxekman: Using default here will cause an error on reload, see https://github.com/webpack/webpack/issues/1685.

justingreenberg: @maxekman no, it will not—require is referencing the default key on imported modules... did you actually merge this PR and encounter an error?

justingreenberg: also fyi, we're using [babel-plugin-add-module-exports](https://github.com/59naga/babel-plugin-add-module-exports) so the default key is only necessary at runtime. if default exists, the plugin attaches a reference directly to module.exports during compilation... for some context, this is the output without plugin:

```js
exports.default = configureStore;
```
output with plugin:
```js
exports.default = configureStore;
module.exports = exports['default'];
```

davezuko:
Awesome, thank you so much @justingreenberg! I'm on the return leg of my trip so finally have a chance to get to all of your PR's.

davezuko:
After some really wonky merge conflicts I finally got this pushed up, thank you for your help as always @justingreenberg!

PR #459, opened 2016-01-15
feat(config) output actual server host

timtyrrell:


davezuko:
Thank you, not sure when this happened :/

PR #461, opened 2016-01-15
chore(react) upgrade react projects to 0.14.6

timtyrrell:


justingreenberg:
fyi this doesn't really do anything—the caret range syntax will resolve the latest release anyway (until v0.15.0 in this case)... if you try a fresh install and run `npm view react version` you'll see 0.14.6

more info: https://github.com/npm/node-semver#caret-ranges-123-025-004

timtyrrell:
Ugh, yes, thanks.

PR #467, opened 2016-01-16
docs(globals): Clarify globals must also be added to .eslintrc, update README

wkw:
- Was not clear to beginner why new webpackDefine globals were not available in source. Add comment to `config/_base.js`
- Update README globals section.

### DIFF
--- a/config/_base.js
+++ b/config/_base.js
@@ -68,6 +68,7 @@ Edit at Your Own Risk
 // ------------------------------------
 // Environment
 // ------------------------------------
+// N.B.: globals added here must _also_ be added to .eslintrc
 config.globals = {
   'process.env'  : {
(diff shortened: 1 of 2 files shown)


config/_base.js
```diff
@@ -68,6 +68,7 @@ Edit at Your Own Risk
 // ------------------------------------
 // Environment
 // ------------------------------------
+// N.B.: globals added here must _also_ be added to .eslintrc
```
justingreenberg: what is N.B.?

wkw: Um, sorry. abbreviation meaning "Note Well"; https://en.wikipedia.org/wiki/Nota_bene. Kind of extra emphasis. I can take it out if it's too obscure.

davezuko:
Thanks so much, I really appreciate your help!

PR #478, opened 2016-01-20
feat(webpack): added koa proxy

yiziz:


davezuko:
Looks good, any way we could just use the `koa-proxy` npm package instead of copying it locally?

yiziz:
@davezuko I made a change here https://github.com/davezuko/react-redux-starter-kit/pull/478/files#diff-4b90ca7d256cf3f4eedad8a25c5ff566R19 that checks `this.originalUrl` instead of `this.path`.  `this.path` is altered by the server middleware and is not an accurate representation of the path the user specified in the browser.  Since I did not see any suitable options to pass into `koa-proxy` that would solve this issue, I copied over the code and made the necessary changes.

davezuko:
Oh wow, somehow never saw that you replied to this. Will try to get to testing this locally ASAP... launch week this week for me, so slower than normal getting to contributions.

davezuko:
Thank you!

PR #481, opened 2016-01-20
add redux logger

plandem:


src/redux/configureStore.js
```diff
@@ -16,7 +16,11 @@ export default function configureStore ({ initialState = {}, history }) {
 
   // Compose final middleware and use devtools in debug environment
   let middleware = applyMiddleware(thunk, routerMiddleware)
-  if (__DEBUG__) middleware = withDevTools(middleware)
+  if (__DEBUG__) {
+    var createLogger = require('redux-logger');
```
thangngoc89: Why `var` ?

plandem:
previous background :) it can be const :)

thangngoc89:
I think it should be `require('redux-logger').default`

davezuko:
CI has been broken for a few weeks now, so going to close this. If you'd like to resubmit with a fixed build and an explanation as to why you think this would be a good addition to the starter kit, I'd be happy to reconsider. Thanks!

plandem:
Sorry, I'm not using react-router anymore, so had to create own starter kit.

PR #488, opened 2016-01-22
Fix sass source maps

web2style:
Source maps were pointing to core.scss if the styles were imported into this file.

### DIFF
--- a/build/webpack.config.js
+++ b/build/webpack.config.js
@@ -46,8 +46,6 @@ webpackConfig.output = {
 // ------------------------------------
 webpackConfig.plugins = [
   new webpack.DefinePlugin(config.globals),
-  new webpack.optimize.OccurrenceOrderPlugin(),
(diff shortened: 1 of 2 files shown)


davezuko:
Thank you!

I had to do some rebase/commit juggling due to some changes to master, but this has been merged (see https://github.com/davezuko/react-redux-starter-kit/commit/b344941241753417fea893fc7a8b46c8b1db8666).

web2style:
I've found out source maps paths are incorrect in this case, because sass-loader adds source maps first, then postcss adds vendor prefixes so the source maps point to wrong lines/files.

I tried to swap sass and postcss, but in this case postcss breaks when parses // comments from sass files. This is weird, because sass loader should have converted them to /* comments. Do these loaders work in parallel? Any way to make these two work synchronously?

Sorry, I'm new to webpack.

PR #489, opened 2016-01-23
Providing user feedback: block the double button on click

eordano:
A feature that is usually needed is the ability to block a button until a request finishes, as users are used to get the feedback that something is going on.

This pull request implements such a feature for the async button. UX looks good, but I have my doubts regarding the implementation.

I feel that this concept should actually be turned into a small library that would wrap a button, set up a "namespace" in the state to handle different buttons, and provide some way to easily incorporate the reducers. I'm new to `redux` development, but I'm fascinated by the mental model. I'm also wary about the dependence between the different actions... looks like it could escalate very quickly to a code mess.

I'd appreciate any feedback that you have on this idea.

Thanks for your work!

### DIFF
--- a/src/redux/modules/counter.js
+++ b/src/redux/modules/counter.js
@@ -1,4 +1,5 @@
 import { createAction, handleActions } from 'redux-actions'
+import { fetch, endFetch } from './fetching'
 
 // ------------------------------------
 // Constants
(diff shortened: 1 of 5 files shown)


rsilvestre:
+1

PR #498, opened 2016-01-27
Load HMR using public rather than relative path.

alexhayes:
I'm using react-redux-starter-kit with django (utilising [django-webpack-loader](https://github.com/owais/django-webpack-loader) and [webpack-bundle-tracker](https://github.com/owais/webpack-bundle-tracker)) and the requests for `__webpack_hmr` fail because they are being attempted relatively.


davezuko:
Thank you!

PR #499, opened 2016-01-27
Change lint to pickup jsx files too

CurtisHumphrey:
flags needed to that jsx files are linted too.  see: https://github.com/eslint/eslint/issues/801

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
   "scripts": {
     "clean": "rm -rf dist",
     "compile": "node -r dotenv/config --harmony bin/compile",
-    "lint": "eslint . ./",
+    "lint": "eslint --est=js --est=jsx . ./",
     "lint:fix": "npm run lint -- --fix",
     "start": "better-npm-run start",
     "dev": "better-npm-run dev",


package.json
```diff
@@ -10,7 +10,7 @@
   "scripts": {
     "clean": "rm -rf dist",
     "compile": "node -r dotenv/config --harmony bin/compile",
-    "lint": "eslint . ./",
+    "lint": "eslint --est=js --est=jsx . ./",
```
ipanasenko: do you mean `--ext` here?

davezuko: @CurtisHumphrey ?

matiit: @CurtisHumphrey 
I am thinking - wouldn't it be better if stuff like that be configurable?

CurtisHumphrey: Hmm, should have been correct syntax but you are right it is --ext not est.

ipanasenko: You should probably fix it then :)

davezuko:
Closing, will gladly accept another PR once the extension has been fixed.Thanks!

Issue #474, opened 2016-01-19
Use bootstrap-sass instead of css


fabiodcorreia:
Hi,

I'm trying to use bootstrap-sass instead of the CDN bootstrap css but I'm having some problems. Here are the steps that I did.

This is probably something easy but it's the first time that I use webpack and I'm probably missing something.

On _base.scss added the following lines

    @import "node_modules/bootstrap-sass/assets/stylesheets/bootstrap-sprockets";
    @import "node_modules/bootstrap-sass/assets/stylesheets/bootstrap";

On core.scss remove the following line since bootstrap-sass already provides normalize and keep the rest as it was.

    @import 'vendor/normalize';

When I run this I get the following error

    Module build failed: CssSyntaxError: postcss-modules-local-by-default:
    (....)/project_name/core.scss:1768:30: Missing whitespace before :global
      background-color: #e6e6e6;
      border-color: #8c8c8c; }

If I remove the :global {} and keep just the code inside the style doesn't get applyed.

Best Regards,
Fábio Correia

seripap:
You have to format those files to be linter-friendly. or just disable the css linter.

fabiodcorreia:
Can I disable the linter just for these files? Format bootstrap files looks a kind of overwhelming.

matiishyn:
Hello @fabiodcorreia 
could you please tell me how to add bootstrap to the project?

I've installed in via npm and have files in `node_modules`

When I'm trying to inject files like you did
`@import 'node_modules/bootstrap/dist/css/bootstrap.min.css';`
then I have an error `Module not found: Error: Cannot resolve module`

When I'm injection like this
`@import '../../node_modules/bootstrap/dist/css/bootstrap.min.css';`
then I don't have an error which means that file is found but still it's not working.

Could you please take a look what I'm doing wrong? Just give me a tip

Thanks in advance

damirv:
Any update on this?

damirv:
I don't have this issue when I place the following code in CoreLayout.js (and bootstrap is being applied correctly):

```
import 'bootstrap-sass/assets/stylesheets/_bootstrap-sprockets.scss'
import 'bootstrap-sass/assets/stylesheets/_bootstrap.scss'
```

fabiodcorreia:
So the problem is just when we add the bootstrap to _base.scss. But shouldn't we add the 3party scss to the _base.scss?

davezuko:
[...]

PR #510, opened 2016-01-29
use bootstrap-loader with bootstrap-sass

damirv:
Possible fix for #474.

@davezuko Not sure if this is what you had in mind, but it worked for me :smile:

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,10 +64,13 @@
   "author": "David Zukowski <david@zuko.me> (http://zuko.me)",
   "license": "MIT",
   "dependencies": {
+    "bootstrap-loader": "^1.0.8",
+    "bootstrap-sass": "^3.3.6",
     "co-request": "^1.0.0",
     "debug": "^2.2.0",
     "history": "^2.0.0",
     "iconv-lite": "^0.4.13",
+    "jquery": "^2.2.0",
     "koa": "^2.0.0-alpha.3",
     "koa-connect-history-api-fallback": "^0.3.0",
     "koa-convert": "^1.2.0",
@@ -79,6 +82,7 @@
     "react-router-redux": "^4.0.0-beta",
     "redux": "^3.0.0",
     "redux-thunk": "^1.0.0",
+    "resolve-url-loader": "^1.4.3",
     "url": "^0.11.0"
   },
   "devDependencies": {
--- a/build/webpack.config.js
+++ b/build/webpack.config.js
@@ -27,8 +27,9 @@ const APP_ENTRY_PATH = paths.base(config.dir_client) + '/main.js'
 
 webpackConfig.entry = {
   app: __DEV__
-    ? [APP_ENTRY_PATH, `webpack-hot-middleware/client?path=${config.compiler_public_path}__webpack_hmr`]
-    : [APP_ENTRY_PATH],
+    ? ['bootstrap-loader', APP_ENTRY_PATH,
+       `webpack-hot-middleware/client?path=${config.compiler_public_path}__webpack_hmr`]
+    : ['bootstrap-loader', APP_ENTRY_PATH],
   vendor: config.compiler_vendor
 }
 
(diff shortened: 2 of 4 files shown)


davezuko:
Hey, thanks for the PR. I'm super swamped with work right now and haven't had much of a chance to sit down and sift through issues/PR's. Will get to this as soon as things settle down a bit.

damirv:
Sure, no problem :)

davezuko:
Just curious, what's the benefit of using bootstrap-loader over a regular import in, say, `core.scss`? Not at all saying it's wrong, just would not have considered that approach.

damirv:
I couldn't get it to work by placing the imports in `core.scss`, I get the following error:
```
ERROR in ./~/css-loader?modules&sourceMap&importLoaders=1&localIdentName=[name]__[local]___[hash:base64:5]!./~/postcss-loader!./~/sass-loader?sourceMap!./src/styles/core.scss
Module build failed: CssSyntaxError: postcss-modules-local-by-default: /Users/damirv/Development/dev/src/styles/core.scss:1:0: Missing whitespace before :global
:global {
```

What i also find weird is the path: `/Users/damirv/Development/dev/src/styles/core.scss`, the actual path is `/Users/damirv/Development/dev/react-redux-starter-kit/src/styles/core.scss` (note the `react-redux-starter-kit` after `/dev/`).

Anyways, the [bootstrap-loader](https://github.com/shakacode/bootstrap-loader) package gives you control of bootstrap styling / component selection through the use of a config file (.bootstraprc), for some this might be an advantage but I can imagine that others want to have manual control. Let me know what your thoughts are on this.

package.json
```diff
@@ -64,10 +64,13 @@
   "author": "David Zukowski <david@zuko.me> (http://zuko.me)",
   "license": "MIT",
   "dependencies": {
+    "bootstrap-loader": "^1.0.8",
+    "bootstrap-sass": "^3.3.6",
     "co-request": "^1.0.0",
     "debug": "^2.2.0",
     "history": "^2.0.0",
     "iconv-lite": "^0.4.13",
+    "jquery": "^2.2.0",
```
davezuko: I really want to pull this in, but this makes me so sad :(

Grobim:
Hi,

I just tried this solution but my app won't start :

bootstrap.js:8 Uncaught Error: Bootstrap's JavaScript requires jQuery
transition.js:59  Uncaught ReferenceError: jQuery is not defined

(Win7, npm : 3.5.3)
(I cloned @damirv repo directly and runned : npm install && npm start)

kevinchiu:
I'm currently using Bootstrap with this kit without any problems. I just put the bootstrap distributable files into src/static/bootstrap/ and added them as normal css/js files in index.html.\:

`
  <script src='/jquery-2.2.0.min.js'></script>
  <script src='/bootstrap/js/bootstrap.min.js'></script>
  <link rel='stylesheet' type='text/css' href='/bootstrap/css/bootstrap.min.css'>
`

davezuko:
Fairly sure I'll just be removing bootstrap from this project altogether and replacing it with a recipe (as well as recipes for `react-toolbox`, `react-materials`, and maybe `semantic`). Really apologize for allowing this PR to linger for so long, I really just wanted to avoid the jQuery dependency and also making this a bootstrap-based project, since it really shouldn't be.

Moving to recipes, thanks to all of you for your contributions and patience.

gomezfx:
@tgfisher Late reply, but maybe this will help someone else. Yeah, using bootstrap-loader, you can setup a .bootstraprc configuration file and then point to a sass file with your customizations.

o1dnik:
@kevinchiu it doesn't solve customization issue with `bootstrap-sass`
@gomezfx but less third-party dependencies makes the project better. I would like to slice and customize in the way I want w/o dependency on bloated `bootstrap`.

package.json
slamdev: what about replacing it by https://github.com/thednp/bootstrap.native

PR #513, opened 2016-02-01
fix(phantomjs): Fix the bug with the rename project of phatomjs to phantomjs-prebuild

rsilvestre:


rsilvestre:
That will fix the bug with phatomjs

PR #529, opened 2016-02-08
deprecate(add-module-exports): remove plugin

davezuko:


GuillaumeCisco:
Hello, I have a lot of problem using `add-module-exports` and was thinking about removing it too, I'd like to know why it is deprecated.
Thank you,

PR #541, opened 2016-02-10
update react-router to 2.0.0, user browserHistory

netpoetica:
in main.js, if you use createHistory, you cannot use [recommended practices](https://github.com/rackt/react-router/blob/latest/docs/guides/NavigatingOutsideOfComponents.md) for transitioning routes via global browserHistory singleton.

Solves issue https://github.com/rackt/react-router/issues/3062

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,7 +69,7 @@
     "react": "^0.14.0",
     "react-dom": "^0.14.0",
     "react-redux": "^4.0.0",
-    "react-router": "^2.0.0-rc5",
+    "react-router": "^2.0.0",
     "react-router-redux": "^3.0.0",
     "redux": "^3.0.0",
     "redux-thunk": "^1.0.0",
--- a/src/main.js
+++ b/src/main.js
@@ -1,13 +1,9 @@
 import React from 'react'
 import ReactDOM from 'react-dom'
-import { useRouterHistory } from 'react-router'
-import { createHistory } from 'history'
+import { browserHistory } from 'react-router'
 import makeRoutes from './routes'
 import Root from './containers/Root'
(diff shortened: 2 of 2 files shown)


davezuko:
Awesome, thank you. Can a base name be configured this way? If not, we'll probably have to add a `<base>` tag to the template.

netpoetica:
Hey @davezuko TBH I am not sure. I have been using the framework for quite a while now and still not sure what the use of the base name ever was? I did a quick search and it looks like other people are starting to question this as well, but it [doesn't look like there is any documentation](https://github.com/rackt/react-router/issues/3045) for it yet. 

[This](https://github.com/MoOx/statinamic/issues/115) might also be a solution, but TBH I can't tell what basename is actually supposed to be.

davezuko:
I've never had to use it myself, but by my limited understanding it's there so that apps running on anything other than the base domain name (thinking an app whose index route is at `http://foo.com/foo`, rather than `http://foo.com`) can have correct relative paths. That said, it can be solved with a tag in the template, so I'll merge this in for now since this is far more important.

daishinlabs:
We also need to update history module `npm install --save react-router history`

davezuko:
@netpoetica also upgraded `react-router-redux` and `history` in the process.

thangngoc89:
Well. For react-router@2 there are 2 ways of doing this :

- use base tag in template
- pull in history package and do it like react-router

```js
import { useRouterHistory } from 'react-router'
import { createBrowserHistory } from 'history'

const browserHistory = useRouterHistory(createBrowserHistory)({ basename: 'foo' })

davezuko:
god how did I miss this, will re-add.

PR #566, opened 2016-02-15
refactor(compile): make webpack compiler reusable

davezuko:


chenfanggm:
@davezuko May I ask, in compile.js, what is ';' in the head of line for?
Thanks

davezuko:
@chenfanggm when writing without semicolons, lines beginning with a `[` or `(` need to be prefixed with a semicolon to terminate the previous expression, since, for example:

```js
const foo = 'bar'
(() => { ... })() // this will try to call 'bar' (think `'bar'(() => { ...})()`)

// or
const foo = 'bar'
[0].toUpperCase()

console.log(foo) // "B" (it accesses the property "0" on `bar` [as in `const foo = 'bar'[0].toUpperCase()`]).
```

chenfanggm:
Thanks :)

PR #569, opened 2016-02-16
fix(config): make Redux actions work with store

jdbence:
```
// Dispatch from anywhere like normal.
store.dispatch(push('/foo'))
```


davezuko:
I intentionally omitted this at first, but it keeps cropping back up as a feature request so looks like we should get it implemented. Thanks for this, will pull it down/merge later!

davezuko:
Merged, thank you!

PR #574, opened 2016-02-17
Made the app link clickable

mmermerkaya:
This small change makes the link clickable in some terminal emulators (like xfce-terminal), which makes it easier to open the app in a browser.


davezuko:
Thanks! Will merge shortly.

davezuko:
Thanks!

PR #576, opened 2016-02-18
Updated eslint to 2.1.0

hamczu:
It works perfectly on npm@3, but temporarily could cause issues on npm@2 because of this bug:
https://github.com/feross/standard/issues/406

### DIFF
--- a/package.json
+++ b/package.json
@@ -103,7 +103,7 @@
     "css-loader": "^0.23.0",
     "cssnano": "^3.3.2",
     "enzyme": "^1.5.0",
-    "eslint": "2.0.0-rc.1",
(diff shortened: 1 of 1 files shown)


davezuko:
Won't we have to wait for babel-eslint and eslint-config-standard to catch up? I haven't dug too much into the new ESLint config changes, but apparently there are some which means existing configs _probably_ won't port flawlessly.

hamczu:
I could install all dependencies and build a project successfully with newer eslint with no errors/warnings.

davezuko:
Rebased your PR and continuing it here: https://github.com/davezuko/react-redux-starter-kit/pull/590

PR #606, opened 2016-02-26
Fixed an issue with using cheerio when chai-enyzme and added tests to prove it

CurtisHumphrey:
So shallow and mount worked but finding elements my class id and render did not until this fix

### DIFF
--- a/build/karma.conf.js
+++ b/build/karma.conf.js
@@ -47,7 +47,6 @@ const karmaConfig = {
     },
     externals: {
       ...webpackConfig.externals,
-      cheerio: 'window',
       'react/lib/ExecutionEnvironment': true,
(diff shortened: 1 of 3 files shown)


davezuko:
Woah, very interesting. Will pull it down when I get a minute. Thank you so much.

davezuko:
I really wish libraries would just include the file extensions, would save so much headache.

Just tested this and it works perfectly, thanks so much for finding it and fixing it.

davezuko:
This has been merged, ignore the close: https://github.com/davezuko/react-redux-starter-kit/commit/12676f6f20116e167725016452da01dce86f6e5e

Thanks again @CurtisHumphrey :D

CurtisHumphrey:
You are welcome @davezuko!

PR #607, opened 2016-02-27
fix(server): use koa-proxy dependency and re-order middleware

davezuko:


server/main.js
```diff
@@ -14,6 +14,11 @@ const debug = _debug('app:server')
 const paths = config.utils_paths
 const app = new Koa()
 
+// Enable koa-proxy if it has been enabled in the config.
```
davezuko: Place this above the `connect-history-api-fallback` middleware so the url is not rewritten.

PR #618, opened 2016-03-01
fix(webpack) Altered config to fix generated resource links

kevinchiu:
We found an issue where generated webpack files app.* and vendor.* were not being linked correctly when deployed to production. Specifying the compiler_public_path in the config fixed the issue for us.


kevinchiu:
It looks like this may have been fixed in another patch.

PR #620, opened 2016-03-01
Redux cli integration

davezuko:


blueprints/blueprint/files/blueprints/__name__/index.js
```diff
@@ -0,0 +1,27 @@
+const path = require('path')
```
davezuko: Fyi @SpencerCDixon this import was missing in the original PR.

blueprints/duck/files/__root__/redux/modules/__name__.js
```diff
@@ -0,0 +1,14 @@
+// Constants
+// export const constants = { }
+
+// Action Creators
+// export const actions = { }
+
+// Reducer
+export const initialState = {}
```
davezuko: Renamed to `initialState` to match current starter kit implementation.

blueprints/dumb/files/__test__/components/__name__.spec.js
```diff
@@ -0,0 +1,8 @@
+import React from 'react'
+import <%= pascalEntityName %> from 'components/<%= pascalEntityName %>/<%= pascalEntityName %>'
```
davezuko: @SpencerCDixon it would be nice to _selectively_ wrap files in folders, since I don't really want ducks/containers/etc. to be wrapped in an extra folder, only components and views.

blueprints/blueprint/files/blueprints/__name__/index.js
SpencerCDixon: Cool thanks.  This is more to just act as like an example page of how to build a blueprint.  Originally I just had everything commented out.  What are you thoughts?  Maybe some more explanation at the top of where to go to find documentation on how to build custom blueprints would be useful?

blueprints/duck/files/__root__/redux/modules/__name__.js
SpencerCDixon: :+1:

blueprints/dumb/files/__test__/components/__name__.spec.js
SpencerCDixon: whoops that's just a remnant of blueprints I use in another project.  The file paths and whether or not they should be wrapped in folders is all determined by how the blueprints are made so that import is actually wrong it shouldnt have the nested structure. 

It should say:
`import <%= pascalEntityName %> from 'components/<%= pascalEntityName %>'` since in the starter kit dumb components are just nested in the /components folder

davezuko: I changed it to nest dumb components in an extra folder, so it is currently correct. I have no updated views yet, though.

.reduxrc
```diff
@@ -0,0 +1,8 @@
+{
+  "sourceBase":"src",
+  "testBase":"tests",
+  "smartPath":"containers",
+  "dumbPath":"components",
+  "fileCasing":"pascal"
```
SpencerCDixon: I choose to use pascal case for filenames by default because it seems like thats what the starter kit uses.  This means if someone does something like: `redux g dumb my-component` the file names will be: `src/components/MyComponent.js`.  If the `fileCasing` setting is set to default then it will use whatever the user enters into the generator for the filename.  I.E. in this example it would of generated `src/components/my-component.js`.  Can you confirm this is the appropriate functionality you want?

blueprints/dumb/files/__test__/components/__name__.spec.js
SpencerCDixon: Oh yeah I see, sweet.  

> no updated views yet

Sorry what do you mean?

davezuko: not* updated views yet (to nest views in a folder similar to components).

SpencerCDixon: > it would be nice to selectively wrap files in folders, since I don't really want ducks/containers/etc. > to be wrapped in an extra folder, only components and views.

just to confirm: you can selectively wrap different blueprints in folders.  Initially I had that as a setting for the CLI but I think using the blueprints model that ember-cli came up with is far superior since it lets each user customize their blueprints and file structures on their own

davezuko: Thinking about it, I totally agree. You can ignore me :)

.reduxrc
davezuko: If you use the default setting, what keyword do you use in the templates to interpolate the filename? Right now it's, for example, `<%= pascalEntityName %>`, is there something like `<%= defaultCase %>`?

davezuko: Good idea, I will take a look at doing that later.

davezuko:
@SpencerCDixon there are a few things I'd still like to cleanup, but this looks solid enough to pull in and get to our users. Anything critical you think is missing, or am I good to merge and then clean it up in smaller pieces over the next couple of days?

SpencerCDixon:
Yeah I'd say its probably good to merge and as bugs crop up we can fix them.  Over time you can adapt the blueprints that are local to this starter kit to be what people agree on being the most useful.  Hopefully eventually people are kind enough to submit PR's of blueprints they have built for personal projects that have been useful!  For example I can imagine one that adds a route to /routes/index. and a corresponding View would be dope :)

davezuko:
Sounds like a  plan.

PR #623, opened 2016-03-02
Changed the naming convention

hamczu:
This is my proposal of better (in my opinion) naming convention. By doing this:
- one doesn't need to specify view/layout name twice in the same path (so it simplifies router and every view when a component is imported)
- one can copy easily import statement for styles - the stylesheet file has always the same name i.e. style.(s)css

I use that convention by myself and I didn't find any drawbacks, just benefits.


src/views/HomeView/index.js
```diff
@@ -3,7 +3,7 @@ import React, { PropTypes } from 'react'
 import { connect } from 'react-redux'
 import { increment, doubleAsync } from '../../redux/modules/counter'
 import DuckImage from './Duck.jpg'
-import classes from './HomeView.scss'
+import classes from './style.scss'
```
davezuko: The reason I avoid this is because it produces less-than-desirable class names w/ CSS modules.

Compare: `HomeView__<class>__<hash>` (changing `HomeView` to whatever component is being used`) vs. `style__<class>__<hash>`.

davezuko:
There have been a handful of issues/PR's revolving around this, and it seems like no matter what we choose it's not going to satisfy everyone. The downside of this approach is that you end up with a bunch of generic `index.js` and `style.scss` files open in your IDE vs. files that are more clearly named.

I will leave this open to give us some time to think about it, but let me go try to dig up some of the old discussions first.

hamczu:
Yep, that's true but in my opinion having the clean and simple code is more important than someone's preferences during file editing. Probably you can can configure IDE to display "better" filename (directory names) in that solution.

I'd totally fine if you don't agree.

src/views/HomeView/index.js
hamczu: I didn't have that issue because I use BEM notation anyway, so inside HomeView/style.css I mainly have .HomeView {} classes

davesnx: Ideally BEM is solving the issue of having the CSS in the global namespace (and alot of more things :P), but It's exactly what you aim with CSS Modules, so... basically don't need to use any notation.

and @davezuko is right, there's no way of do a convention that likes to everyone! :baby_chick:

hamczu: Yep, I know that BEM was created to solve global CSS namespace, but for me it's just cleaner to use it even with CSS modules. Because of CSS modules I usually drop .Element--modifier convention and just add class .modifier

davezuko:
Just got back from vacation; really sorry for the delay.

Going to close this for now as we'll be sticking with the current conventions (see https://github.com/SpencerCDixon/redux-cli/issues/17#issuecomment-192422141 for more discussion), but @nathanielks has some good ideas for making this style more accessible, so look for that in his PR's and upcoming changes.

Thanks so much for the contribution though, I really appreciate it and the discussion.

davezuko:
See: https://github.com/davezuko/react-redux-starter-kit/commit/4e21062642c13be978ba7fa8c49ecc4c571aa509

PR #630, opened 2016-03-06
Added reducer blueprint

ericwooley:
The reducers are what I use the blueprints for the most so far, figured I would contribute back my blueprint for reducers.

If you don't want a reducer blueprint, feel free to reject.

### DIFF
--- a/blueprints/reducer/index.js
+++ b/blueprints/reducer/index.js
@@ -1,0 +1,27 @@
+const path = require('path')
+
+module.exports = {
+  locals: function (options) {
+    // Return custom template variables here.
+    return {}
+  },
+
+  fileMapTokens: function (options) {
+    // Return custom tokens to be replaced in your files
+    return {
+      __token__: function (options) {
+        // logic to determine value goes here
+        return 'value'
+      }
+    }
+  },
+
+  // Should probably never need to be overriden
+
+  filesPath: function () {
+    return path.join(this.path, 'files')
+  },
+
+  beforeInstall: function (options) {},
+  afterInstall: function (options) {}
+}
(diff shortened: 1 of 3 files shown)


blueprints/reducer/index.js
```diff
@@ -0,0 +1,27 @@
...
+
+  // Should probably never need to be overriden
+
+  filesPath: function () {
+    return path.join(this.path, 'files')
+  },
+
+  beforeInstall: function (options) {},
+  afterInstall: function (options) {}
+}
```
SpencerCDixon: I would remove the boilerplate in here and the path require

ericwooley: by path require, you mean line 21 - 23?

SpencerCDixon: line 1 of blueprints/reducers/index.js.  You can have it just export an empty object since you don't need to override any of the hooks.

I think there may already be a `duck` blueprint that does something like this, but might be useful to have different versions of blueprints.

davezuko:
Generic but non-automatic response: away on vacation, ty for PR, will review next Saturday/Sunday.

davezuko:
We already have a duck reducer (naming is a bit unclear for the uninitiated), so is this needed?

ericwooley:
@davezuko I didn't realize the duck was supposed to be a reducer blueprint. And it doesn't follow the other blueprint, but meh! do as you like!

davezuko:
I'll consider it further, but will probably drop one or the other.

davezuko:
This will be changing with https://github.com/davezuko/react-redux-starter-kit/pull/684. Thanks for the contribution though, much appreciated!

PR #632, opened 2016-03-07
fix(eslint): pin eslint to ~2.2.0 to avoid missing estraverse-fb

### LINKED ISSUE #629: estraverse-fb missing
I just downloaded, installed and attempted to run 2.0.0-alpha.1.  "npm start" complained that estraverse-fb was missing.  I did a save-dev install of estraverse-fb, and now all appears to be fine.  I see no errors in the "npm install" log, and no references to estraverse-fb.

neverfox:
Fixes #629

### DIFF
--- a/package.json
+++ b/package.json
@@ -102,7 +102,7 @@
     "css-loader": "^0.23.0",
     "cssnano": "^3.3.2",
     "enzyme": "^2.0.0",
-    "eslint": "^2.1.0",
+    "eslint": "~2.2.0",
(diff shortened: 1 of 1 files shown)


davezuko:
Unable to read through issues in much detail atm, but I don't think this should be solved at the project level iirc since it's a known issue with one of the dependencies iirc.

davezuko:
Apologies, missed some info in my email and didn't see that you fixed it by locking eslint. Will merge as a hotfix while I am away. Thank you.

PR #633, opened 2016-03-07
added missing module

npancha2:
Added missing module estraverse-fb to package.json.


davezuko:
Off on vacation so unable to look closely, but iirc this should be fixed at least the dependency level, since it's not a concern of this project.

davezuko:
Hotfixed by locking eslint with @neverfox's PR. Thank you though!

PR #634, opened 2016-03-07
Feature/update blueprints

nathanielks:
Okie doke! Per [this comment](https://github.com/SpencerCDixon/redux-cli/issues/17#issuecomment-192492537) I've updated how the different entities are stored and referenced. I've updated the blueprints to generate the forms/layouts/views as well as updated the existing ones to follow the same convention:
```
entity/{name}/{name}Entity.js
entity/{name}/index.js
```


nathanielks:
@davezuko What pattern would you like to follow: node style (eg `import HomeView from 'views/Home')` or the more explicit (eg `import HomeView from 'views/Home/HomeView'`)? I'll push another update per your desires!

davezuko:
Generic but non-automatic response: away on vacation, ty for PR, will review next Saturday/Sunday.

davezuko:
Looks great @nathanielks; the only thing I'm not on board with is the shortening of the folder names (`HomeView` -> `Home`, `CoreLayout` -> `Core`) since this goes against the current structure of the project. I understand it's personal preference, so I'd like to leave it up to users to update the blueprints themselves if they prefer your style.

If you could update those I'd love to get this merged in. Thanks.

nathanielks:
Will do! What about forms and their folder? Currently they're just files. Do you want to keep that convention or put them in a directory?

davezuko:
I'd prefer everything be nested inside of a directory for consistency, and that way it's easy for users to modify blueprints to their liking since everything is the same. If there are blueprints _not_ currently following this convention you can assume it's a mistake.

nathanielks:
Okie smokie, @davezuko! This should be good to go :+1:

davezuko:
Awesome, thank you. Rebased/squashed these locally and merged, so PR won't auto-close, but this has been merged into master: https://github.com/davezuko/react-redux-starter-kit/commit/4e21062642c13be978ba7fa8c49ecc4c571aa509.

nathanielks:
Ha, I was just about to ask if you wanted me to squash :joy:

PR #659, opened 2016-03-18
fix(prod-config): use absolute paths in webpack's output

beeftornado:
Building the project with the production configuration outputs an html file with the js and css linked as relative urls and if you have subpaths in your app, then anyone arriving at a subpath first will fail to find the resources as it will look for them in that subpath on the server.


davezuko:
Thanks for the fix, I appreciate it.

PR #675, opened 2016-03-30
Remove strip_root from .flow_config in order to keep Nuclide happy

nottombrown:
Out-of-the-box nuclide appears to choke on the `strip_root` flag


davezuko:
Thanks for the report. Willing to remove it temporarily if it's temporarily causing issues, but I feel like this is better suited for a bug report in Nuclide, since `strip_root` is a [valid flow configuration option](http://flowtype.org/docs/advanced-configuration.html).

davezuko:
Had to rebase against master so git doesn't recognize the PR as merged, but it's [in master](https://github.com/davezuko/react-redux-starter-kit/commit/0af2d0daa1871ea63a460a814bdc2c749b4a602d). Thanks!

Issue #662, opened 2016-03-22
Production dependencies vs dev dependencies


nhagen:
Wondering what the line is between dev dependencies and production dependencies here? For any kind of deployment, we need to have installed many of the devDependencies like babel and eslint, no?

davezuko:
It is pretty silly right now, since of course every dependency is required in order to build the project, and it isn't like a distributed library where only certain dependencies need to be installed to use it vs. develop on it.

I tried to make the distinction along the lines of "this is what's needed to run the application (React, Redux, etc.)" vs. "this is what's needed to _build/develop_ the application (Babel, etc.)." As you point out, this isn't exactly a valid approach, but since you should be installing devDeps anyways if you're building the project it should be a non-issue. Because of this, I personally liked separating the two just for clarity of what is needed in the app for production builds vs. everything else.

nhagen:
I can agree with your reasoning. I will say though for practical reasons, for example if you're trying to optimize build time (including `npm install --production`), that having the minimum possible set of dependencies for production would be nice, so for example keeping only develop modules like `redbox-react`, `redux-devtools`, etc., in `devDependencies`. That doesn't cut out a ton of time really, but our dev ops team was slightly confused about why we were getting so many `module not found` errors for things that were obviously needed for building in production. It was because our make file was adding the `--production` flag to npm.

davezuko:
That's valid, but it would require a few tweaks to the build system (i.e. we would have to only use eslint-loader when in development). It's totally feasible, so I'll consider shuffling things around; however, I am curious then about how your team handles testing (is the environment w/ `--production` different than your CI environment for testing? Because if not, then things like Karma, PhantomJS, and the rest would all have to go in deps as well).

nhagen:
We move most things from `devDependencies` to `dependencies` and leaving some select things out. Initially before we started testing this made a lot of sense because we only needed to move a few deps. With testing obviously you need to move almost everything.

Its just when we copy over new stuff from this repo that we have to comb through all of the `devDependencies` and decide what is what over again. Not a real easy way to merge `package.json` in our case.

davezuko:
Ok, I will revisit the ordering of the dependencies this weekend and figure out something that makes a bit more sense.

davezuko:
[...]

PR #681, opened 2016-03-31
refactor(dependencies): move production dependencies to regular deps

davezuko:
Resolves https://github.com/davezuko/react-redux-starter-kit/issues/662


davezuko:
Resolves https://github.com/davezuko/react-redux-starter-kit/issues/662

PR #682, opened 2016-03-31
Feature/Self Contained Apps (SCAs)

alexhayes:
This is something that I've been doing for a while now and it's not uncommon to the way a lot of frameworks are structured, for good reason. As outlined in the README;

> The idea of a Self Contained App (SCA) is that you can keep everything specific to an section of your project within the one directory and not have pieces spread out all over the file system. This partitioning is similar to the concept of an app in [Django](https://docs.djangoproject.com/en/1.9/intro/tutorial01/#creating-the-polls-app) (among others).
> 
> There are a number of advantages to using SCAs;
> 
> - Less spread of naturally grouped items across the project file system.
- Easier to [code split](https://webpack.github.io/docs/code-splitting.html) the project.
> - Refactoring becomes much easier because everything is contained in the SCA.
> - Import using relative paths within the SCA rather than nested folders from the project root.
> 
> ```
...
...
├── src                      # Project source code
|   ├── apps                 # Split your project up into Self Contained Apps
|   │   └── home             # Home app
|   |       ├── components   # Home app React Components (generally Dumb components)
|   |       ├── views        # Home app components that live at a route
|   |       ├── tests        # Home app unit tests
|   |       ├── reducers.js  # Home app Redux reducer
|   |       └── routes.js    # Home app ReactRouter Routes
...
...
```

Happy to squash down to a single commit if desired but my intention was to open this so that the feature could be discussed with possible changes.


alexhayes:
And obviously I'll fix up the tests :)

justingreenberg:
this is great—i'm working on something very similar. i will open a PR later tonight when i have a few minutes... @davezuko any initial reaction to this concept?

davezuko:
@alexhayes this is great, thanks for the contribution! To contain the discussion, I suggest we all contribute to https://github.com/davezuko/react-redux-starter-kit/pull/684.

PR #683, opened 2016-04-01
fix(travis): error in travis. "npm run deploy" set node_env to production by default

rsilvestre:
by creating "npm run deploy:no-env", we don't set node_env to production by default and let the user (travis) change this env from dev to prod to test all the deployment type


davezuko:
Wow, I can't believe this slipped by for so long (granted it probably cropped back up after the re-introduction of better-npm-run). Thank you!

davezuko:
Also, going to just make this the default (`no :no-env`) and add helpers for dev and staging.

davezuko:
Squashed the extra merge commits and made some tweaks, check out https://github.com/davezuko/react-redux-starter-kit/pull/689. Your original commit is still in there, thanks again for catching this :tada: .

PR #685, opened 2016-04-01
'/' path for 'compiler_public_path' has been added.

softage0:
'/' path for 'compiler_public_path' has been added since compiled assets are unable to be loaded when the param-included route URL directly accessed.

This change adds '/' in front of URL that presents the location of complied assets in 'index.html' in 'dist' directory.

### DIFF
--- a/config/_base.js
+++ b/config/_base.js
@@ -30,7 +30,7 @@ const config = {
   compiler_hash_type       : 'hash',
   compiler_fail_on_warning : false,
   compiler_quiet           : false,
-  compiler_public_path     : '',
+  compiler_public_path     : '/',
   compiler_stats           : {
     chunks : false,
     chunkModules : false,


davezuko:
I feel like there was a reason for leaving this blank originally... let me think about it for a little but, and if I can't figure out a counter argument we can get this merged. Thanks.

softage0:
Yep, it could be technically the issue of `react-router` or `react-router-redux`.
My solution might be makeshift way :)
Hope you find a counter argument and I think it must be resolved since this issue could be encountered often.

MarcusOhman91:
+1, changed this in my project a week ago already and haven't stumbled upon any issues

davezuko:
This has been merged: https://github.com/davezuko/react-redux-starter-kit/commit/17fd5c2ee5042581d9b7fc5676ca0bbe444b4838.

PR #689, opened 2016-04-02
Fix/deploy env

davezuko:


rsilvestre:
Smart, it's better like that

Issue #697, opened 2016-04-06
Can't run test with fbjs 0.8.0


Freddy03h:
## Issue
Hello, I installed Draft.js on my project, a _Rich Text Editor by Facebook_ : https://facebook.github.io/draft-js/
And this project has the particularity to be dependent of **fbjs@0.8.0-alpha.2**

And since, test doesn't run anymore, when I `npm run test`, I've got the following error : 
```
START:
06 04 2016 17:53:21.039:INFO [karma]: Karma v0.13.22 server started at http://localhost:9876/
06 04 2016 17:53:21.071:INFO [launcher]: Starting browser PhantomJS
06 04 2016 17:53:22.960:INFO [PhantomJS 2.1.1 (Mac OS X 0.0.0)]: Connected on socket /#gba2WCoQqyELMJKhAAAA with id 86581400
PhantomJS 2.1.1 (Mac OS X 0.0.0) ERROR
  TypeError: undefined is not a constructor (evaluating 'query.startsWith(name)')
  at /Users/freddy/Desktop/react-redux-starter-kit/tests/test-bundler.js:91609 <- webpack:///~/draft-js/~/fbjs/lib/UserAgent.js:35:0

Finished in 0.654 secs / 0 secs

SUMMARY:
✔ 0 tests completed
```
## How to reproduce
You can reproduce it on a new project, after cloning the starter kit : `npm install draft-js --save` and add dependancies of this lib to a view, for example add the following code to `src/views/HomeView/HomeView.js` : 
```
import {
  Editor,
  EditorState,
  RichUtils,
  Entity,
  CompositeDecorator,
} from 'draft-js'
```
and `npm run test`

## related ?
I don't know if it's an issue related to this project, or directly related to fbjs or maybe phantomjs. So, sorry if it's not related to this project.

I also noticed that react@15.0.0 will use fbjs@0.8.0 https://github.com/facebook/react/blob/master/package.json#L38
So it will be also problematic for the react update.

Freddy03h:
Ok, thank's to the great help of @zpao on this issue https://github.com/facebook/fbjs/issues/133#event-617440260

I fixed the issue by adding `babel-polyfill` to the `karma.conf.js` file.

davezuko:
[...]

PR #698, opened 2016-04-06
Use babel-polyfill instead of phantomjs-polyfill in karma.conf

Freddy03h:
Phantom doesn't include some new method in ES6. We need to include a polyfill for some of these things in karma.conf file for running test.

Related issue : https://github.com/davezuko/react-redux-starter-kit/issues/697

### DIFF
--- a/build/karma.conf.js
+++ b/build/karma.conf.js
@@ -9,7 +9,7 @@ debug('Create configuration.')
 const karmaConfig = {
   basePath: '../', // project root in relation to bin/karma.js
   files: [
-    './node_modules/phantomjs-polyfill/bind-polyfill.js',
+    './node_modules/babel-polyfill/dist/polyfill.js',
     {
       pattern: `./${config.dir_test}/test-bundler.js`,
       watched: false,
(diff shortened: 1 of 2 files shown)


davezuko:
Thanks for the PR. There is a note in the README with [information about babel-polyfill](https://github.com/davezuko/react-redux-starter-kit#babel-polyfill), but perhaps it would be better to just include it. If we _do_ end up including it, I'd like to include it in both the main bundle _and_ the test suite so the environments match.

davezuko:
Will be merging this, just have a few other things lined up first. :tada:

Freddy03h:
I also think we should add `babel-polyfill` to the webpack config for the main bundle, as you say.

And maybe add also a fetch polyfill ?

davezuko:
Still considering this... I can honestly see it going either way.

Freddy03h:
Do you want me to update the PR to add `babel-polyfill` in the webpack config ?

davezuko:
@Freddy03h sure, can you add it to the webpack entry point in the default bundler as well as in the karma configuration? Then we can get this merged.

davezuko:
Implemented with https://github.com/davezuko/react-redux-starter-kit/commit/fedfd79f0795cb0aa561424d28ce3334be3e878b. Had to add it to the app since we were getting a bunch of issues from people not understanding problems in IE.

PR #710, opened 2016-04-14
Nitrous.io Quickstart button

yeouchien:
Added Nitrous Quickstart files so anyone can quickly spin up a free, cloud dev environment to start using this repo. Clicking the Quickstart button will automatically clone this repo and will install and configure any required packages.


davezuko:
This looks awesome and I appreciate the contribution, but I don't really want to go down this road since it will inevitably lead to similar configurations for Heroku and whatnot. I _will_, however, add a note to direct people to your PR so they can see how easy it is if they choose to use your service.

PR #712, opened 2016-04-14
feat(eslint-loader): added optional npm script to lint during development

jflayhart:
For some of us not running monolithic React apps and using this boilerplate, we would like to keep the option to run eslint during live reloads. It's optional and added as a separate script, if desired. It's scalable so if a project does get too large to run eslint while webpack compiles, then you can run the other script. Easier than changing the webpack config each time, I think.

### DIFF
--- a/build/webpack.config.js
+++ b/build/webpack.config.js
@@ -96,21 +96,21 @@ times for larger projects. `npm run lint` still exists to aid in
 deploy processes (such as with CI), and it's recommended that you
 use a linting plugin for your IDE in place of this loader.
 
-If you do wish to continue using the loader, you can uncomment
-the code below and run `npm i --save-dev eslint-loader`. This code
-will be removed in a future release.
-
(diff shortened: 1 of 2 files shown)


jflayhart:
This was inspired by [HelplightningXeon](https://github.com/HelplightningXeon) on my front-end team. We have gratefully been using this boilerplate, with some modifications, to kick off a project re-write we are doing.

davezuko:
See: https://github.com/davezuko/react-redux-starter-kit/issues/713

jflayhart:
I think it's wise to be consistent and not flip flop your deprecations, so take or leave this PR. It was very easy to do and I figured I may as well submit a PR in case others find this helpful.

davezuko:
Closing due to my own inattention, which I am incredibly sorry about. I really appreciate the contribution but am going to maintain the decision to keep the loader out of the base project.

PR #720, opened 2016-04-16
Adding this will not show those annoying npm errors at the end after linting

codetony25:
adding the || true after the eslint . will not show the npm errors at the end and keep them silent. It's annoying to see in my opinion. :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
   "scripts": {
     "clean": "rimraf dist",
     "compile": "better-npm-run compile",
-    "lint": "eslint .",
+    "lint": "eslint . || true",
     "lint:fix": "npm run lint -- --fix",
     "start": "better-npm-run start",
     "dev": "better-npm-run dev",


davezuko:
I'm not familiar with this syntax outside of JS. Is this compatible on both Windows and Unix systems? Sadly Travis does not also run a windows box.

kid:
I think this will also make the Travis build report false success in case of lint errors no?

If you want to avoid theses errors, you can do `npm run -s lint` or `npm run --silent lint`

davezuko:
@kid that's what I thought as well.

tonytainguyen:
yeah, --silent works as well. However having --slient means that the user has to type in npm run lint --silent flag everytime when they want to lint. The reason for me putting "|| true" in the end was to avoid the user having to do that. @davezuko you brought up if this works on windows, I have not tested this on windows, but when I have time, i'll try it out on my friends computer.

marshallford:
@codetony25 This seems like a good idea however it could break scripts (like Travis) that rely on exit codes. As a test I added `npm run lint` to my pre-commit git hook and noticed git was still committing after warnings displayed when I appended `|| true` to my eslint command. However, it looks like using the silent flag in the git hook solves both issues (as in, doesn't commit and doesn't show npm error blob).

davezuko:
I appreciate the sentiment but I'd rather just put this in the FAQ without possibly breaking CI build processes. Will add a note to myself to refer people to your solution. Thanks!

PR #753, opened 2016-04-26
Webpack: parallellize builds

justingreenberg:
So, [**`HappyPack`**](https://github.com/amireh/happypack) is something i've been meaning to dig into for a while—it spawns multiple loader instances and when properly tuned can decrease build times to a fraction of what they were before. i've been beating this up for a few days, and everything is working great. the effect on build time is negligible on the starter kit, but the returns on large projects are huge. @davezuko please let me know what you think! some benchmarks at bottom.....

Todo: 
 - Fix CI
 - Add config documentation to README
 - add happypack cache to npm script `clean` (rimraf) 

#### Some technical details from README:

HappyPack sits between webpack and your primary source files (like JS sources) where the bulk of loader transformations happen. Every time webpack resolves a module, HappyPack will take it and all its dependencies, find out if they need to be compiled[1], and if they do, it distributes those files to multiple worker "threads".

A diagram showing the flow between HappyPack's components

Those threads are actually simple node processes that invoke your transformer. When the compiled version is retrieved, HappyPack serves it to its loader and eventually your chunk.

[1] When HappyPack successfully compiles a source file, it keeps track of its mtime so that it can re-use it on successive builds if the contents have not changed. This is a fast and somewhat reliable approach, and definitely much faster than re-applying the transformers on every build.

#### Benchmarks

For the main repository I tested on, which had around 3067 modules, the build time went down from 39 seconds to a whopping ~10 seconds when there was yet no
cache. Successive builds now take between 6 and 7 seconds.

Here's a rundown of the various states the build was performed in:

Elapsed (ms) | Happy?  | Cache enabled? | Cache present? | Using DLLs? |
------------ | ------- | -------------- | -------------- | ----------- |
39851        | NO      | N/A            | N/A            | NO          |
37393        | NO      | N/A            | N/A            | YES         |
14605        | YES     | NO             | N/A            | NO          |
13925        | YES     | YES            | NO             | NO          |
11877        | YES     | YES            | YES            | NO          |
9228         | YES     | NO             | N/A            | YES         |
9597         | YES     | YES            | NO             | YES         |
6975         | YES     | YES            | YES            | YES         |

-

DLLs are coming next :)


config/_base.js
```diff
@@ -44,6 +44,13 @@ const config = {
     'react-router-redux',
     'redux'
   ],
+  // Parallellization! via HappyPack
+  thread: true,
```
justingreenberg: thread config is here

package.json
```diff
@@ -121,7 +122,7 @@
     "sass-loader": "^3.0.0",
     "style-loader": "^0.13.0",
     "url-loader": "^0.5.6",
-    "webpack": "^1.12.14",
+    "webpack": "^1.13.0",
```
justingreenberg: update to 1.13.0

davezuko:
If those build times are real...

colbert-jaw-drop

davezuko:
So I just went and applied these changes to our largest production app (will have to get an official LoC count) and didn't see any change in build times (even on subsequent compiles without any code changes). I'll keep fiddling with it throughout the day, as if this _does_ offer such tremendous improvements I'd love to get them in, but I'd like to figure out exactly what tweaks are needed to realize them.

justingreenberg:
@davezuko did you try tweaking the threadpool count in config? 

another thing to try... for the longest build (cssModules maybe?), try replacing the shared threadPool reference in Thread config with `threads: 4` and let the happypack plugin internally manage it's own threads. see if scaling the threads up and down has any impact...

>This number indicates how many Node VMs HappyPack will spawn for compiling the source files. After a lot of tinkering, I found 4 to yield the best results. There's certainly a diminishing return on this value and increasing beyond 8 actually slowed things down for me.

> Keep in mind that this is only relevant when performing the initial build as HappyPack will switch into a synchronous mode afterwards (i.e. in watch mode.) Also, if we're using the cache and the compiled versions are indeed cached, the threads will be idle.

davezuko:
@justingreenberg I tried a few (3, 4, 5) and they were all about the same. I wanted to just get a quick idea of what immediate impact this had, but I'll keep at it to see if I can't get it to have some effect (it was almost the exact same as pre-happypack, so I'm guessing there's something I may be missing).

davezuko:
One thing I realized that was different between this project and the app I'm testing on is that it has eslint-loader as a pre-loader, so I figured that that could have influenced the build time. However, after disabling it, I actually had _faster_ build times without happy pack than with it (even with subsequent builds). Other than that, the webpack configurations are nearly identical.

justingreenberg:
Yeah something must be off. There are too many people posting evidence of their builds dropping from 22s to 10s, etc. I gotta look into this more tonight. My gut tells me there's something going on with our sass/postcss config...

Possibly related: https://github.com/amireh/happypack/issues/14#issuecomment-208254692

justingreenberg:
sorry to let this get stale i've been absolutely swamped... i'm going to close this for now, will take another stab at this later in the week

davezuko:
@justingreenberg no need to close, I just haven't gotten it to work yet and have been a bit too busy to mess with it much further. Up to you.

PR #754, opened 2016-04-26
Update routes and docs

justingreenberg:
Implements and documents some of my learnings...

 - Update README
 - Make base routes synchronous and add explanation
 - Add module name for weback async chunks

 ```sh
#  ...instead of module id
1.1.a400dddf35fd91e3e653.js

#  outputs module name...
1.counter.5eef484d990f2a85bb5b.js
 ```
 - Detailed inline comments on:
    - How and where to create split points, specifying what to bundle
    - What does webpack do/what does react-router do, how do they integrate
    - Sync vs Async route definitions and their relationship with code splitting


davezuko:
Nice, will get this in later today if I can.

justingreenberg:
sounds good, i wanted to add a small note about integrating plain route definitions with JSX (based off #755) i'll push something up now

davezuko:
Had to squash some of the sync commits, but this is in master. Thanks.

PR #758, opened 2016-04-28
feat(webpack): allow user to serve app over local network

### LINKED ISSUE #757: hardcoded "http://localhost:3000" prevents preview on other device (e.g. phone)
I tried to preview a work-in-progress on my phone today (running `npm start`). The initial HTML request is returned as expected, however, the script and stylesheet URLs all have "http://localhost:3000" hardcoded into them, which means there's only one machine that can successfully load it: the same one running `npm start`.

Is there an option I have missed to control this? Any suggestions? Cheers!

P.S. I'm super excited about this template. Thanks so much for the great work so far. :)

marshallford:
Fixes #757. I also resolved the linting errors. The [ip](https://www.npmjs.com/package/ip) module also supports the host being offline, i.e. if you aren't connected to a network, it will fallback to localhost.


jokeyrhyme:
This would work just fine for my purposes.

The only downside I can see is that it's all or nothing: all clients have to connect via an IP address, or all clients have to connect via "localhost". It'd be interesting to know what happens if the `npm start` machine has multiple NICs with different IP addresses, too.

build/webpack.config.js
```diff
@@ -154,7 +154,7 @@ const PATHS_TO_TREAT_AS_CSS_MODULES = [
 // If config has CSS modules enabled, treat this project's styles as CSS modules.
 if (config.compiler_css_modules) {
   PATHS_TO_TREAT_AS_CSS_MODULES.push(
-    paths.client().replace(/[\^\$\.\*\+\-\?\=\!\:\|\\\/\(\)\[\]\{\}\,]/g, '\\$&')
+    paths.client().replace(/[\^\$\.\*\+\-\?\|\\\/\(\)\[\]\{\}]/g, '\\$&')
```
davezuko: What's the rationale for this change?

marshallford: It was marked as a linting error until I made the suggested changes. If you want to keep it as is, I would recommend an eslint ignore line comment. If you aren't getting a linting error it could be an issue with my text editor as I have standard js installed as seperate package along with eslint related packages.

jokeyrhyme: Yeah, the ESLint error is that these characters don't need to be escaped with a back-slash.

jokeyrhyme:
I can confirm that this PR allows the `npm start` web server to be accessed via

- localhost:3000 from the same device

- 192.168.x.x:3000 from a difference device (IP address varies based on your network settings)

Great work. :)

Is it worth rebasing this on master to clear up the merge conflicts?

marshallford:
@jokeyrhyme Good to hear! As for rebasing, to be honest I am not familiar with that git feature so I am not sure how to proceed at the moment.

jokeyrhyme:
When I work on a PR, I usually have my git branches set up something like this:
- master: tracking upstream "master", this project
- local-ip: tracking "local-ip" branch in my own fork

Which this setup, I can do the following:

1. `git checkout master`
2. `git pull`
3. `git checkout local-ip`
4. `git rebase master`
    - this will have conflicts, so you'll need to sort them out, use `git add ...` to mark them as resolved, and `git rebase --continue` until you are all done
5. `git push -f` (be super careful about this, make sure you are in your own branch, tracking your fork)

After this, the changes in the PR are as though you forked today, ready to merge without any conflicts.

marshallford:
@jokeyrhyme Thanks! I will look at this in the morning. Follow up question: I have my fork git cloned locally (master and local-ip now out of date), how would I set master to pull from a different repo (in this case davezuko/react-redux-starter-kit) instead of my fork?

jokeyrhyme:
You can use the [`git remote`](https://git-scm.com/docs/git-remote) command to set up the remotes. For example, I usually have the following setup:
- origin: the upstream project, this project
- fork: my fork of the project

Then you can use the [`git branch`](https://git-scm.com/docs/git-branch) command to configure your branches.

Assuming you have "origin" and "fork" remotes, you could do something like:

- `git branch --set-upstream-to=origin/master master`
- `git branch --set-upstream-to=fork/local-ip local-ip`

From there you should be able to `git pull`, etc.

davezuko:
I can rebase it as well, you will just see this PR get "closed" rather than "merged", even though your commit(s) will show up in master. I will pull this down today to give it a go.

davezuko:
Merged, thank you! Sorry for the turnaround time on this.

PR #767, opened 2016-05-02
feat(blueprint): Add route blueprint

danhuynhdev:
Make blueprint for the new route structure


SpencerCDixon:
So we're going to need to make a decision how to handle where to store smart/dumb/container components and my guess is those blueprints will also need to be refactored.

One idea: we could pass a flag to the generator like `-r <route name>` and if that flag is there we generate the component in that route's Smart/Dumb/Container component folder.  Implementing that might be somewhat tricky but I could look into.  WIthout the flag the components could just get generated in their normal Smart/Dumb/DuckModule locations.  Although, I get the feeling when using the Fractal structure there is no longer the normal paths since all components/redux should be nested in their specific routes.

Lemme know what you guys think

justingreenberg:
i think that's a good idea—maybe the default behavior could output "common" containers/containers to root and the -r param could provide nesting

danhuynhdev:
I will try to implement that kind of behavior. And fix the duck generator.

danhuynhdev:
But where do we put old duck modules. I don't see a place in Dave's structure tree where i can put it. The store folder may be a good place but Dave say we shouldn't reflect the store there?

ethanve:
Also, where do you plan on putting redux-forms?

danhuynhdev:
@SpencerCDixon I have to rely on a hack in index.js because filesPath doesn't have options pass into it. Could you do some thing on this.

SpencerCDixon:
Instead of relying filesPath can you create a custom __token__ for fractal files?  If the flag isn't there we replace that __fractal__ token with an empty string so components get rendered in the fallback location.  If the flag is there we replace __fractal__ with the proper -r RouteName.  

You can also add logic/normalization/prompts to users in the generation of the new token.  [This comment thread might help to understand what I'm talking about.](https://github.com/SpencerCDixon/redux-cli/issues/70#event-649566750)

`filesPath` should only be used if you wanted to put blueprints in a folder OTHER than `blueprints`.  An example use case of this is in the ember-cli they have this concept of addons and addons can come with their own blueprints so that option is available to extend the cli to offer other locations for blueprints.

danhuynhdev:
I remove all the hack.
And here are example usage.
- redux g route name
```
src/routes/Name
├── components
│   └── Name.js
├── containers
│   └── NameContainer.js
├── index.js
└── modules
    └── name.js
tests/routes/Name
├── components
│   └── Name.spec.js
├── containers
│   └── NameContainer.spec.js
├── index.spec.js
└── modules
    └── name.spec.js
```
- redux g dumb name -r Name
```
src/routes/Name
└── components
    └── Name
        ├── index.js
        └── Name.js
tests/routes/Name
└── components
    └── Name
        └── Name.spec.js
```
- redux g dumb name
```
src/components/Name
├── index.js
└── Name.js
tests/components/Name
└── Name.spec.js
```
- redux g smart name -r Name
```
src/routes/Name
├── components
│   └── Name
│       ├── index.js
│       └── Name.js
└── containers
    └── NameContainer.js
tests/routes/Name
├── components
│   └── Name
│       └── Name.spec.js
└── containers
    └── NameContainer.spec.js
```

davezuko:
👍  thank you for the PR. I will be away for most of this weekend but I look forward to taking a look at this next week. If I don't reply feel free to comment back to give me a nudge :P.

blueprints/smart/files/__root__/__routepath__/__route__/__dumb__/__name__/__name__.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react'
+
+type Props = {
```
cluk3: What is the purpose of the `type` keyword here? Could not find anything on Google about it.

davezuko: It is flow and I did recently remove it just to slim things down. We'll have to adjust accordingly.

cluk3:
A flag to add .scss file and import it in the component could be also useful, what do you think?

davezuko:
I haven't had a chance to sit down and fully test this out yet, so I apologize for the delay.

davezuko:
Eliminating blueprints in an effort to simplify this project. I appreciate the contribution, however. Thanks!

Lube:
removing blueprints was a bad call imo, any issue to discuss this?

davezuko:
Blueprints are great, but I really don't think they're vital to what comes in this project out of this box. I've felt sort of disillusioned lately with how much is going on in this project, so I'm picking things back up by trying to slim down the stack as much as possible. A full documentation rewrite will be coming with that, and if blueprints don't make it back in (which they very well might [make it back in]) then they will at least receive a decent writeup in the installation section.

I'd be happy discussing it here, even though it is in a random PR, or in the gitter channel. I _may_ open up a "What the hell is going on?" issue that we can dump all these into, but not convinced that's necessary just yet.

Lube:
This boilerplate has helped me kickstart and understand react/redux, so I'm very grateful for all your efforts, I will contribute with my opinion here and wherever you decide to centralize the discussion.

I do indeed think that the complexity that the blueprints add is easily tackled with documentation, once you figure out how they work, its impossible not to take advantage of them and learn how build some for your own personal use.

Thanks again for coming back and I'll hope to keep contributing to the repo.

PR #822, opened 2016-05-23
Feature/setup

raksonibs:
Inspired by react-generator tasks, I thought these similar tasks could be useful!


davezuko:
I've been away from this project for a while, so I apologize for the ridiculously delayed response. I'm currently working to simplify this project and remove as much code as possible, so while this is definitely a great idea, I'm going to have to decline just to keep us moving in the right direction.

Thank you for your time, I appreciate the contribution.

PR #828, opened 2016-05-26
Replacing redux-thunk with rxjs 5 and redux-observable

nathanvale:
Thanks to Jay Phelps and Ben Lesh, redux-observable is middleware for redux that is inspired by redux-thunk.

[https://medium.com/@benlesh/redux-observable-ec0b00d2eb52#.icqm622pm](url)

_Redux-observable allows developers to dispatch a function that returns an observable, promise or iterable of action(s). When the observable emits an action, or the promise resolves an action, or the iterable gives an action out, that action is then dispatched as usual._

I think that this is a fantastic way to bring powerful and fun asynchronous programming into the react and redux libraries! Some people might by interested in using RXJS 5 with redux so I thought I would share with your great starter kit.

I have refactored the tests/routes/Counter/modules/counter.spec.js spec to show how easy it is to write tests for async actions that can be chained and even cancelled! Promises can still be used with this middleware which leads you to consider needing redux thunk at all.


justingreenberg:
i doubt this will make it into the kit, but this is a great PR 👍 awesome work

andyhu:
+1 redux-thunk and redux-saga are both too hacky. really look forward for this patch to be merged.

nathanvale:
Great news is that redux-obserable is replacing its thunk style approach which means that redux-observable and think-middleware can sit along side each other. Unknown this kits is meant to be simple but showing a reactive programming approach would really help other developers.

https://github.com/redux-observable/redux-observable/issues/44

src/routes/Counter/modules/counter.js
```diff
@@ -47,7 +47,7 @@ const ACTION_HANDLERS = {
 // ------------------------------------
 // Reducer
 // ------------------------------------
-const initialState = 0
+const initialState = 5
```
jarohmh: Why you changed initialState?

nathanvale: Because from a UX perspective clicking on Double Async before incrementing appears to do nothing - you cant double 0!

kolpav:
Well there are many things from react/redux ecosystem which are really great but I still think they have no place in this project. The thing which made this starter kit so popular in my opinion is it managed to keep itself super minimal and it serves as great first step into react world. You have bare minimum to increment number by clicking on a button even that looks like a magic for a newbie. The moment you start replacing simple things like `redux-thunk` even when they have its faults with something more sophisticated like `redux-observable`,`redux-sagas` you just made life much harder for newcomers. Sooner or later when they start to build their own applications they might discover that thunk is simply not enough for them. I still think this PR is great maybe there could be **recipes**/**built with this start kit** section somewhere.

src/routes/Counter/modules/counter.js
jarohmh: Maybe some comment for this line then?

andyhu:
In that case, maybe it's better to create a PR to https://github.com/mxstbr/react-boilerplate, which main focus is best practices

src/store/createStore.js
```diff
@@ -1,15 +1,15 @@
 import { applyMiddleware, compose, createStore } from 'redux'
 import { routerMiddleware } from 'react-router-redux'
-import thunk from 'redux-thunk'
+import {reduxObservable} from 'redux-observable'
```
andyhu: Are you sure it's  `{reduxObservable}` rather than reduxObservable?

nathanvale: I have seen createEpicMiddleware and looking at updating this example as thunkservables are not deprecated.

nathanvale:
thunkservables are now deprecated. this approach is now dated.

PR #859, opened 2016-06-13
[WIP] Exhanced documentation in README

ralphschindler:
This is a work in progress to fill in some gaps in the README.  Personally, I'd like to see the README be a place that gives you enough direction to know where and how to start adding code to your react-redux-starter-kit based project.  Now that the project has been moved to fractal, and some of the code resources have as well.. I have a few questions that would help me better understand where things are intended to go.  And, if worthwhile, I'll add any information to the README so that others can benefit.

Questions:

1. Currently, counter's presentational component is located at `src/components/Counter/*`, while its reducer and component wiring (container) is located at `src/routes/Counter/*`.  Conversely, the current skeleton (tree in README) demonstrates that `src/routes/Counter/components` is an ideal place to put react components that are highly related to the route they live at. Which should it be? In the Counter example, the component living in the `routes` seems more ideal since it is less "reusable" than other kinds of components that might exist at `src/components`.  (For example, just a thought... if there was a CounterWidget that many pages could share, whereas the Counter component might be the presentation that lives at the route of /counter).

2. The redux directory is currently left out of the tree view.  I think it should remain, and or potentially be renamed from `src/redux` to just `src/modules`.  It seems that location is designed for "ducks" (which I believe should be linked from the README, or briefly explained).  If there is something else that would exist at `src/redux` but not be inside `src/redux/modules`, I think that should be explained as well.

3. Currently, using `redux-cli` and the blueprints puts forms at `src/forms`.  Is this ideal? Should forms be promoted to a top-level thing? It seems as though they potentially live in `src/components` or in `src/routes/*/components`. Thoughts?

4. Should "forms" be smart, ie: should they have their own `containers` folder so that they can have a container that does the reduxForm() as well as the mapping of properties and passing of actions to be reduxified?

Thanks for the great project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,15 +50,47 @@ Finally, This project wouldn't be possible without the help of our many contribu
 
 ## Getting Started
 
-After confirming that your development environment meets the specified [requirements](#requirements), you can follow these steps to get the project up and running:
+After confirming that your development environment meets the specified [requirements](#requirements), you can create a new project based on `react-redux-starter-kit` in one of two ways:
+
+### Install from source
+
+First, clone or download:
 
 ```bash
 $ git clone https://github.com/davezuko/react-redux-starter-kit.git
-$ cd react-redux-starter-kit
+// or
+$ wget -O react-redux-starter-kit.zip https://github.com/davezuko/react-redux-starter-kit/archive/master.zip
+$ unzip react-redux-starter-kit.zip
+```
+
+Then, rename to your project name and change into the directory:
+
+```bash
+$ mv react-redux-starter-kit <my-project-name>
+$ cd <my-project-name>
+```
+
+### Alternatively, install via `redux-cli`
+
+If not already installed (globally):
+
+```bash
+$ npm i redux-cli -g
+```
+
+Then, create a new project:
+
+```bash
+$ redux new <my-project-name>
+$ cd <my-project-name>
+```
+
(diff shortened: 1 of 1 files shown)


davezuko:
👍 Much appreciated. Wrapping up the work week and will get to this ASAP.

_Please_ bug me if this goes untouched for too long.

ralphschindler:
We should probably circle around on this in the next week, if you are available.  I can join any particular chat if you're in any of them.

davezuko:
Hey, going to merge this in, and then we can move on with those questions you asked. I'm awfully sorry this sat around for so long.

ralphschindler:
Thanks. I'll try to sum up the remainder.

Essentially, the inclusion of fractal and blueprint give us a good starting point, but with more "tools" than before.  Previously, there were only a few locations one would expect to start writing new (JS) code: `views/` (for things routed to), `components/` (Dan's concept of dumb react components), and `containers/` (context providing components... admittedly, didn't get much use out of this directory).

Now we have more named directories with fractal, and each has their own sub-directory structure. Essentially, it would be nice to be able to have some kind of guide that as I write new code, I can isolate the best possible place to put it according to both a nice set of examples and some specific spelled out in the README guidelines. (And further, when to split concepts up across multiple files, that might have been all in one previously.)

In short, I just need to know which "tool" to employ as I build an app.

PR #866, opened 2016-06-16
Update Heroku deployment section

woniesong92:
I just went through deploying it to Heroku and decided to document it for other people. @davezuko let me know if you want to make some changes or feel free to reject this ;)

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,7 +145,47 @@ Out of the box, this starter kit is deployable by serving the `~/dist` folder ge
 If you are serving the application via a web server such as nginx, make sure to direct incoming routes to the root `~/dist/index.html` file and let react-router take care of the rest. The Koa server that comes with the starter kit is able to be extended to serve as an API or whatever else you need, but that's entirely up to you.
 
 ### Heroku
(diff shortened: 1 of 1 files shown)


davezuko:
This is great, I don't see why I should reject it ;)

woniesong92:
@davezuko ooh or feel free to merge it!

jsfr:
One thing though. Note that doing the heroku setup in this way, you are doing the webpack build during the startup of a dyno. This means the webpack build runs every time a dyno is spinned up/restarted. It also means you only have 60 second for the build to finish and the app to bind to the port assigned by heroku as per https://devcenter.heroku.com/articles/error-codes#r10-boot-timeout

So the second you start building anything but a pretty trivial app, you probably want to fork and maintain your own version of https://github.com/heroku/heroku-buildpack-nodejs to use webpack durings herokus build phase instead of the startup/deploy phase. This also means you get 15 mins to build instead, and that heroku builds on faster machines when you start using anything but hobby instances.

An example of a  buildpack with webpack is here https://github.com/jsfr/heroku-buildpack-nodejs (it simply runs an extra npm task)

Info on buildpacks can be found here https://devcenter.heroku.com/articles/buildpacks

davezuko:
Merged, thank you!

davezuko:
@jsfr I saw you commented on this but can't seem to find it. If you have any concerns with what was just merged, please let me know and I'd be happy to edit. I'm _not_ a Heroku user, so I'm taking this more on faith and what little experience I have with it than rigorous experience.

PR #868, opened 2016-06-17
koa-static 3.x now supports koa2, so we don't need to convert anymore

### LINKED ISSUE #864: Use koa-static v3.0.0
koa-static already supports koa2, so we don't have to use convert for it. Why you don't migrated into it yet? Is there any reason for that?

chompomonim:
Fix #864


davezuko:
Thank you :)

PR #876, opened 2016-06-21
fix(store): fix hot reloading not working with root reducers

rochdev:


davezuko:
Totally thought this had already been merged, ugh. Thanks!

PR #883, opened 2016-06-25
Created fractal pattern blueprints, closes #734

### LINKED ISSUE #734: Update CLI generators to work with Fractal Structure
/cc @justingreenberg @SpencerCDixon

ericwooley:
The blueprint for a route is a large blueprint, perhaps it could be broken into several?

In any case, this is better than the outdated ones.


davezuko:
Looks good, but I don't think putting the duck image in the blueprints is necessary. If you'd like to remove that from the commit we can merge this in!

ericwooley:
@davezuko Deleted!

davezuko:
Merged!

PR #884, opened 2016-06-25
Test fix

kolpav:
I am just getting started with enzyme but I think `Should render exactly two buttons` test doesn't make sense in its current state. What do you think?


davezuko:
You're absolutely right. Do you think you could squash your commits and reword the message to follow our [guidelines](https://github.com/davezuko/react-redux-starter-kit/blob/master/CONTRIBUTING.md)? If I do it your PR won't automatically close, so up to you.

kolpav:
Next time I will read the guidelines:) is it ok now?

davezuko:
Works for me, thank you!

PR #893, opened 2016-06-28
feat(react-hot-loader): begin migration

davezuko:


src/main.js
```diff
@@ -40,37 +40,30 @@ if (__DEBUG__) {
 // ========================================================
 const MOUNT_NODE = document.getElementById('root')
 
-let render = (routerKey = null) => {
-  const routes = require('./routes/index').default(store)
+// TODO(zuko): get rid of routerKey
+// TODO(zuko): hot reloading does actually work, the whole app reloads
```
davezuko: doesn't**

UnsungHero97:
@davezuko what happened with the `react-hot-loader` feature? this was closed, but it doesn't seem like the migration actually happened. was this closed because `react-hot-loader` won't be adopted?

davezuko:
@UnsungHero97 I closed it because react-hot-loader is still in beta and this PR has been super long running. Rather than letting it sit stale, I figured it was best to close it and then re-evaluate once react-hot-loader v3 stable is released.

UnsungHero97:
thanks for responding! cool! that makes sense.

PR #918, opened 2016-07-09
Fix(karma): Fix Proper incremental test running.

### LINKED ISSUE #871: Hot/Incremental testing is currently broken
This is not a major issue, but I just wanted to share my findings since I spent some time today debugging why hot testing is not working anymore.

How to reproduce:
1. Check out the source
2. Run `npm run test:dev`
3. Make a change to a .spec file
4. Expect that only the tests for the .spec file are executed

This is a regression as this feature used to work before. Here's what happened:

This starter kit ships with https://github.com/aaronjensen/karma-webpack which enables hot testing. https://github.com/davezuko/react-redux-starter-kit/issues/572 had a problem where the build was broken and it was fixed by this commit https://github.com/davezuko/react-redux-starter-kit/pull/709/commits/e5ca664780070750c81e4df362ea1074ce4cead9.

This commit replaces `const __karmaWebpackManifest__ = []` with `const __karmaWebpackManifest__ = new Array()` in test-bundler.js. This makes the tests compile again but at the same breaks hot testing because of the way karma-webpack works. karma-webpack actually parses the file and string replaces `__karmaWebpackManifest__ = []` with the list of files to test (see https://github.com/aaronjensen/karma-webpack/blob/master/index.js#L241).

So in order to get hot testing to work we need to revert the change but need to figure out why it breaks the build. I can actually reproduce that `npm test` fails with `__karmaWebpackManifest__ = []` on my machine (OSX, node 5.10.1, npm 3.8.3).

````
20 06 2016 17:10:39.173:INFO [PhantomJS 2.1.1 (Mac OS X 0.0.0)]: Connected on socket /#W0RkVPTegbppEO_0AAAA with id 81782802
PhantomJS 2.1.1 (Mac OS X 0.0.0) ERROR
  SyntaxError: Expected token ']'
  at /Users/drichard/src/react-redux-starter-kit/tests/test-bundler.js:55 <- webpack:///tests/test-bundler.js:6:0
````

What's interesting is that `npm run test:dev` works without any issues. What does test:dev do differently? It doesn't calculate code coverage.

Sure enough if we disable coverage for `npm test` (for example by removing the isparta loader in karma.conf.js) the tests run just fine.

I'm not entirely sure what's happening here but it seems that both karma-webpack and isparta modify the source file in a way that breaks it (have a look at the compiled output: https://gist.github.com/drichard/14362488a37f67a48b5569b1eee5994f).

In short:
1. code coverage + aaron jensen's karma-webpack fork can break the build
2. The workaround disabled hot testing
3. If you want hot testing, remove the work around and and turn off code coverage (until we figure out why it's broken in the first place)

gerbal:
Resolves #871 

Declaring `__karmaWebpackManifest__` as a `new Arrray()` breaks incremental the incremental test runner. As it prevents `__karmaWebpackManifest__` from updating with as the webpack manifest changes. 

The ugly fast fix is to use an array literal and terminate the statement with a ';' to avoid regressing #709

### DIFF
--- a/build/karma.conf.js
+++ b/build/karma.conf.js
@@ -68,7 +68,7 @@ if (config.globals.__COVERAGE__) {
     test: /\.(js|jsx)$/,
     include: new RegExp(config.dir_client),
     loader: 'isparta',
-    exclude: /node_modules/
+    exclude: [/node_modules/, /test-bundler/]
   }]
 }
 
--- a/tests/test-bundler.js
+++ b/tests/test-bundler.js
@@ -21,10 +21,9 @@ global.should = chai.should()
 // Require Tests
 // ---------------------------------------
 // for use with karma-webpack-with-fast-source-maps
-// NOTE: `new Array()` is used rather than an array literal since
-// for some reason an array literal without a trailing `;` causes
-// some build environments to fail.
-const __karmaWebpackManifest__ = new Array() // eslint-disable-line
+// NOTE: A trailing `;` is used because its absense breaks some test environments
+// `new Array()` breaks the `__karmaWebpackManifest__` variable. 
+const __karmaWebpackManifest__ = []; // eslint-disable-line
 const inManifest = (path) => ~__karmaWebpackManifest__.indexOf(path)
 
 // require all `tests/**/*.spec.js`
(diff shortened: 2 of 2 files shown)


gerbal:
After further experimentation and exploration, the solution to this one is really easy. Simply excluding `test-bundler` from test coverage appears to resolve the issue.

davezuko:
Looks like this got fixed in another PR (the `Array` -> `[]` fix). Not sure the `exclude` is necessary since we have an explicit `include`, but if I'm wrong in that please let me know and I can rebase this and merge that fix.

Thanks!

Issue #845, opened 2016-06-05
troubleshooting action creators


brandonmp:
I'd normally throw a question like this on stackoverflow, but I suspect the problem is something I'm missing in the structure of redux elements across files in this project.

I am creating a component that calls Firebase's Google login popup.  After writing the action creators & component, login worked fine. However, after trying to write the reducers, the action creators (`thunks`, all) aren't getting past their respective `return` statements. 

I've tried reverting the changes, but I'm pretty new & not awesome w/ VCS. 

so: Is there a straightforward way to debug a situation like this? is there another file i should've made changes to?

a few details 
- the store object from the `duck` initializes w/ initial state
- `console.log()` statements work if they're in an action creator, but before the return statement. if they're in a `thunk`, nada

(and pls. feel free to point out ill-advised patterns or dumb stuff i'm doing. always looking to get better)

// containers/LoginWidget.jsx (mistakenly combined container/presentation component, plan on splitting once it's working)

```js
import React from 'react';
import Paper from 'material-ui/Paper';
import RaisedButton from 'material-ui/RaisedButton';
import {attemptLogin, logUserOut} from 'containers/LoginWidget/modules/user-auth.js';
import {connect} from 'react-redux';

class LoginWidget extends React.Component {

  constructor(props) {
    super(props);
  }


  render() {
    return (
      <Paper style={styles.wrapper}>
        <RaisedButton
          label="Sign in with Google"
          labelStyle={styles.loginButton}
          backgroundColor={'#D84B37'}
          icon={<img height={15} src={require('./assets/google.png')}/>}
          name="google"
          onTouchTap={() => this.props.dispatch(attemptLogin)}
        />
        <RaisedButton
          label="Logout"
          labelStyle={styles.loginButton}
          backgroundColor={'#D84B37'}
          name="logout"
          onTouchTap={() => this.props.dispatch(logUserOut)}
        />
      </Paper>
    )
  }
}

export default connect()(LoginWidget);
```

// containers/LoginWidget/modules/user-auth.js
```js
// ------------------------------------
// Constants
// ------------------------------------

//user authentication statuses
export const USER_NOT_AUTHENTICATED = 'USER_NOT_AUTHENTICATED'; //ie, just arrived to site & hasn't logged in yet
export const USER_AUTH_IN_PROGRESS = 'USER_AUTH_IN_PROGRESS';
export const USER_LOGGED_IN = 'USER_LOGGED_IN';
[...]

PR #936, opened 2016-07-17
fix(blueprint): simplify actions boilerplate

oori:
1. Reduce the boilerplate surrounding adding actions.   with this patch,  you only need to add actions in the module, not in the container.
2. Updated variable name to better reflect redux naming.   see: https://github.com/davezuko/react-redux-starter-kit/issues/845#issuecomment-223837496


davezuko:
Blueprints are being removed in an effort to simplify this project. I appreciate the PR, however. Thank you for your contribution!

PR #941, opened 2016-07-20
chore(core): Remove random key from render

justingreenberg:
Using the random key hack for react router has been keeping me up at night, this is just a much more explicit approach...

```js
ReactDOM.unmountComponentAtNode(MOUNT_NODE)
```

😄


davezuko:
Hey there! Thanks for the fix, I've been pretty distant from this project lately so this is much appreciated. I cherry picked your commit into master because of the duplicate commit history in the PR, but we should be good to go: https://github.com/davezuko/react-redux-starter-kit/commit/a602625b20ade172c829b583882ae44c150e738c.

PR #943, opened 2016-07-20
Add redux-logger package. 

unclechao:


marshallford:
Is this essential? This PR would make for a great wiki entry but I don't see the need for this dependency  when the [Redux DevTools](https://chrome.google.com/webstore/detail/redux-devtools/lmhkpmbekcpmknklioeibfkpmmfibljd?hl=en) Chrome extension already adds logging and time travel.

PR #948, opened 2016-07-23
ESLint configuration

anthonyraymond:
Added Intellij config folder to `.gitignore`
Added already isntalled plugins to `.eslintrc`


davezuko:
Nice changes, thank you!

Issue #976, opened 2016-08-22
Add createReducer Abstraction


low-ghost:
Would be useful to have an abstraction for the pattern of creating a reducer from an object where the key corresponds to an action type. Based on [Redux docs](http://redux.js.org/docs/recipes/ReducingBoilerplate.html#reducers). Since this is probably the inspirational source, perhaps there's a good reason for not including this. Too much magic? Also, establishes pattern for application wide utility functions and (not included) the pattern can be applied to fractal routes. 'Utils' vs 'helpers' semantics... Templating via redux-cli alleviates the manual repetition, but I'm for pulling out code that will be used in every reducer module such as this.

Mentioned in [#934](https://github.com/davezuko/react-redux-starter-kit/issues/934), and probably elsewhere.

Also, awesome project.

low-ghost:
PR #977

davezuko:
We did use a reducer/action creator abstraction in the past, but I really wanted to get away from bombarding users with additional concepts/abstractions/dependencies to understand. The argument could honestly go either way, since one side argues for demonstrating good practices by avoiding repetition, while the other prioritizes an understanding of the fundamentals.

low-ghost:
No problem, figured that was the case. It's a level of abstraction I like, especially if I'm going to be building things out quickly w/ redux-cli. But, I'll just keep my branch in those cases. Thanks

davezuko:
It is definitely a very useful abstraction to have, and one I personally use as well. Unfortunately, one thing we have to consider is that a lot of people picking up this project are new to this ecosystem as a whole, so I try to limit abstractions where possible to get to the meat of the project. My hope is that other users will organically figure some of this stuff out, learn what questions to ask, and possibly find their way back to this very issue :).

PR #977, opened 2016-08-22
feat(redux): Add createReducer Abstraction

low-ghost:
Pr adds an abstraction for the pattern of creating a reducer from an object where the key corresponds to an action type.

[Issue #976](https://github.com/davezuko/react-redux-starter-kit/issues/976)


davezuko:
I appreciate the contribution, but we used to have this and I removed it in the effort of eliminating indirection and keeping things as simple as project.

PR #980, opened 2016-08-23
fix(compile): report errors in webpack compilation

chieffancypants:
Errors in configuration were getting swallowed and reporting `TypeError: Cannot read property 'toJson' of undefined` instead of the actual root cause.  This is because there is an attempt to read the `stats` object (undefined in the case of a compile error) before checking the `err` object


davezuko:
Nice, thank you!

davezuko:
This has been merged: https://github.com/davezuko/react-redux-starter-kit/commit/81b2dfa576291dd55e986f0536dfb4ea661de74b. I had to rebase it, so the PR doesn't reflect its true status. Thank you again.

PR #988, opened 2016-09-01
chore(package.json): deploy runs linter and test

maaaikoool:
According to the README.md:
```
npm run deploy : Runs linter, tests, and then, on success, compiles your application to disk.
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
       }
     },
     "deploy": {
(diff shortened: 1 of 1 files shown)


davezuko:
Good catch, not sure when that changed, or if it was just a documentation error (though I believe it to be the former).

davezuko:
Merged, thank you!

PR #989, opened 2016-09-04
style(eslint): Enforce linting on all code

neverfox:
I cannot think of a good reason why all code shouldn't come under the same linting rules, so this expands `npm run lint` to cover all major files, including tests.


davezuko:
Agreed, I can't remember why I would have excluded tests, so I always assumed they were linted.

PR #990, opened 2016-09-04
upgrade(multiple): Upgrade of all dependencies that are safe to upgrade

neverfox:
This is a bit dangerous in that it's sweeping, but it should be a safe, across-the-board upgrade. It excludes `history` and `cheerio` to avoid incompatibilities.


davezuko:
Most of these upgrades fall within the `^` rules, so they'd be installed anyways, but I don't see the harm. Thanks.

davezuko:
Rebased and merged, PR status is incorrect.

davezuko:
(Thank you!)

PR #991, opened 2016-09-04
Pull request/eslint rules

neverfox:
This is subjective (and includes changes from `pull-request/linting`) but if it's acceptable, I think it's an improvement. No worries if not.


.eslintrc
```diff
@@ -20,10 +20,21 @@
...
+      },
+      "multiLine" : {
+        "align"       : "colon",
+        "beforeColon" : true,
+        "afterColon"  : true
+      }
+    }],
+    "babel/generator-star-spacing" : 1,
+    "generator-star-spacing"       : 0,
+    "jsx-quotes"                   : [2, "prefer-double"],
```
davezuko: I explicitly overrode StandardJS's double quote preference in JSX, but the rest look good.

build/webpack.config.js
```diff
@@ -7,18 +7,18 @@ import _debug from 'debug'
...
 
 debug('Create configuration.')
 const webpackConfig = {
-  name: 'client',
-  target: 'web',
-  devtool: config.compiler_devtool,
-  resolve: {
-    root: paths.client(),
-    extensions: ['', '.js', '.jsx', '.json']
+  name    : 'client',
```
davezuko: it's so beautiful, just like the early days of this project.

.eslintrc
neverfox: I can revert that one no problem.

davezuko: If you wouldn't mind :).

neverfox: Done

davezuko:
👍

pleunv:
> The primary goal of this project is to remain as unopinionated as possible

I'm not really sure why changes like these are being merged in?
I'm trying to maintain and sync a few forks that we use as boilerplates for our own projects, and (opinionated) changes to linters (and all affected files) are really messing this up, especially if these get reverted back a few days later. I know that a boilerplate is supposed to be a kickstart for a new project, however with some care you should be able to keep a branch with modifications on top of the boilerplate updated fairly easily.

Perhaps my rebasing strategy is bad, however rebasing our modifications on top of the master branch every few weeks feels like the best approach to do this. I'm not sure if there's any other approach that could somehow alleviate some of the nightmare that comes with having to rebase on top of a drastically changed linter config?

I'm open for suggestions, but right now it feels like this might not be sustainable for us very much longer. We will either have to freeze to an older master commit or look at other more easily maintainable/updatable solutions.

Would separating build scripts, dummy project setup, ... into separate modules be a viable approach? Or is that not a direction you want to take with this boilerplate?

PR #992, opened 2016-09-07
ESLINT with Airbnb

chintan9:
just update your repo.


Dr1Ku:
You included your SSH private key here, wrong origin when pushing?

davezuko:
Closing this because of the SSH key being exposed, you will want to remove that commit from your remote so people can't find it.

PR #1000, opened 2016-09-12
Simplify Project

davezuko:


chompomonim:
Wow, you should migrated from koa into express, deprecated css-modules and blueprints. Any motivation on each of this changes?

davezuko:
@chompomonim from the gitter channel:

```
A lot of [the changes] are opinionated, and I'm sure many will disagree. The team I work for at my daytime job is currently working on eliminating babel from our stack, and it's really felt like a breath of fresh air. Removing it plus, on the client, the multiple transforms between the source code and what finally gets executed has helped debugability tremendously and makes it easier to share code without having to make sure everything has a compatible build system. It's become more of a "let's stop always pushing for new all the time, even if it sounds awesome, and instead leverage what we have natively and only import the things that solve our specific needs."

The koa -> express is just purely to eliminate the need to transpile async/await functions, which is sad because koa is brilliant. I know generators could still be abused with co, but ultimately it's easier to just stick with express, especially for such a small server. It does suck to have CommonJS on the server and ES6 imports on the client, but it's great not having to transpile the Node server. And, since Node supports a lot of other ES6 features, most of the other code does not look that different. It does hamper universal rendering, but that's out of this project's scope anyways.

My thoughts [about this project] going forward are... I'd really love to just get everybody using the same thing, and that doesn't have to be my project. My personal goal is to start more thoroughly what we believe to be mandatory to an application. Whenever I used this project to start a new one I'd always find myself stripping out a ton of stuff, which I eventually realized indicated some sort of flaw.
```

terencechow:
Why remove css modules? Seems like it removes functionality that promotes good css practices.

davezuko:
@terencechow I may add them back in with very basic configuration and disabled by default. A large percentage of the issues received in this project related to CSS modules: importing other packages, how to use them, wondering why styles aren't hot reloaded, tricky errors with sass compilation, etc. This is fine, but I don't have the time to support these numbers of issues, so I'd rather give people something they're more familiar with -- and then, per your comment, possibly the option of deciding for themselves if they want to invest their time into CSS modules.

Removing them entirely may not have been necessary, so I will reconsider, but by default I think plain old sass is sufficient if you stick to good naming and practices.

cryptoquick:
I hate to be the pitchforks here, and your progressiveness is admirable, but this really is too soon. For me, this project stopped useful development at `4e53e68`. Guess I'll check back in 2018, when browserside async/await becomes widespread...

davezuko:
That's fine, and people are totally entitled to their opinions. Ultimately this project stopped being useful to me because of how progressive it was trying to be and just how much _stuff_ was going on, so I made the understandably controversial decision to rework it. I apologize (no sarcasm) if this makes your life more difficult or means this project is no longer a viable option to you. I'd love nothing more than to use all of the features supported by Babel but not yet by Node, but ultimately I've come to value the simplicity and peace of mind gained by eliminating transpilation.

hirako2000:
I used this starter kit at both of those stages (with Koa, then with Express).
I find the decision totally justified, beyond the simplification arguments. Folks use a starter kit to get familiar with a given stack - having too many sensitive parts is not always helping to figure out each library.

Koa is awesome, but dragged me into hours of troubleshooting because it's so cutting edge (changing APIs, more scarce solid doc/example than Express). I even had to wrap my own multer as for multi-part forms to get exactly what I needed.

There is plenty to learn on the React/Redux side of things - server can be swapped with Koa, and add css modules for the client side as one goes along with building an app.

Thanks for this project, one of the best starter kit out there.

PR #1003, opened 2016-09-14
fix(nodemon): nodemon mustn't exec node to start the server

akkie:
The command `nodemon --exec node bin/server` does not start the server. Instead it shows the wrong usage of the `nodemon` command. Moreover, nodemon is a node wrapper, so it's not needed to call node as external program.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
       }
     },
     "dev": {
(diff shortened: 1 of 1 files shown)


davezuko:
Ah fudge, my mistake. Thank you for catching this.

PR #1005, opened 2016-09-18
fix(home component): remove implied use of CSS modules (deprecated)

ecjs:
Awesome starter kit..love the structure!  This tripped me up a bit at first, because I assumed CSS modules were being used..it looks like you just had some remnants of its use before it was deprecated from the webpack config.


davezuko:
Ah good catch, you are exactly right. Thank you.

neverfox:
The kit dropped CSS Modules?

volodymyr-tsaryk:
@ecjs could you please explain why css modules were deprecated ?

PR #1013, opened 2016-09-21
feat(core): init store with sync location reducer

### LINKED ISSUE #1011: Warning: Store does not have a valid reducer.
Hi
Without modifying any code I get this warning in the browser 
Store does not have a valid reducer. Make sure the argument passed to combineReducers is an object whose values are reducers

The counter example is working though.

justingreenberg:
cleaned up #1012 — resolves #1011 with simple `location` reducer (broken out into separate module and tested) that effectively keeps location in sync with store and works with HMR

known limitations...
- **will not** replay in address bar during time travel debugging
- **will not** respond to `LOCATION_CHANGE` events, other than to update the state... history actions should be made using `browserHistory` singleton or ie using `withRouter` HOC

these limitations can be mitigated using middleware, if necessary

### DIFF
--- a/src/containers/AppContainer.js
+++ b/src/containers/AppContainer.js
@@ -1,21 +1,24 @@
 import React, { Component, PropTypes } from 'react'
-import { Router } from 'react-router'
+import { browserHistory, Router } from 'react-router'
 import { Provider } from 'react-redux'
 
 class AppContainer extends Component {
   static propTypes = {
-    history : PropTypes.object.isRequired,
-    routes  : PropTypes.object.isRequired,
-    store   : PropTypes.object.isRequired
+    routes : PropTypes.object.isRequired,
+    store  : PropTypes.object.isRequired
+  }
+
+  shouldComponentUpdate () {
+    return false
   }
 
   render () {
-    const { history, routes, store } = this.props
+    const { routes, store } = this.props
 
     return (
       <Provider store={store}>
         <div style={{ height: '100%' }}>
-          <Router history={history} children={routes} />
+          <Router history={browserHistory} children={routes} />
         </div>
       </Provider>
     )
--- a/src/main.js
+++ b/src/main.js
@@ -1,11 +1,10 @@
 import React from 'react'
 import ReactDOM from 'react-dom'
-import { browserHistory } from 'react-router'
 import createStore from './store/createStore'
 import AppContainer from './containers/AppContainer'
 
 // ========================================================
-// Store and History Instantiation
+// Store Instantiation
 // ========================================================
 const initialState = window.___INITIAL_STATE__
 const store = createStore(initialState)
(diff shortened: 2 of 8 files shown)


src/containers/AppContainer.js
```diff
@@ -1,21 +1,24 @@
...
 class AppContainer extends Component {
   static propTypes = {
-    history : PropTypes.object.isRequired,
-    routes  : PropTypes.object.isRequired,
-    store   : PropTypes.object.isRequired
+    routes : PropTypes.object.isRequired,
+    store  : PropTypes.object.isRequired
+  }
+
+  shouldComponentUpdate () {
```
justingreenberg: no need to ever re-render this component

src/main.js
```diff
@@ -58,12 +53,12 @@ if (__DEV__) {
     }
 
     // Setup hot module replacement
-    module.hot.accept('./routes/index', () => {
-      setTimeout(() => {
+    module.hot.accept('./routes/index', () =>
+      setImmediate(() => {
```
justingreenberg: tldr setImmediate uses IO queue for scheduling (as opposed to setTimeout using call stack) so it is similar to `process.nextTick` in node and thus more performant

src/store/createStore.js
```diff
@@ -32,6 +34,9 @@ export default (initialState = {}) => {
   )
   store.asyncReducers = {}
 
+  // To unsubscribe, invoke `store.unsubscribeHistory()` anytime
+  store.unsubscribeHistory = browserHistory.listen(updateLocation(store))
```
justingreenberg: assign a reference to store so subscription is cancellable

src/store/location.js
```diff
@@ -0,0 +1,31 @@
+// ------------------------------------
```
justingreenberg: tried to make this as similar to other redux modules/code style as possible

tests/test-bundler.js
```diff
@@ -32,6 +32,6 @@ const testsToRun = testsContext.keys().filter(inManifest)
 
 // require all `src/**/*.js` except for `main.js` (for isparta coverage reporting)
 if (__COVERAGE__) {
-  const componentsContext = require.context('../src/', true, /^((?!main).)*\.js$/)
+  const componentsContext = require.context('../src/', true, /^((?!main|reducers).)*\.js$/)
```
davezuko: If this was just to get codecov to be happy it's not necessary, since I can just merge anyways.

davezuko:
🎉  Thank you @justingreenberg :).

justingreenberg:
@davezuko no worries! hopefully it does without saying but please don't feel any pressure to keep any of these changes—please update/revert/restructure to your liking :) i experimented with a few setups, ie using AppContainer lifecycle for managing history subscriptions would fire changes on HMR, and the location reducer didn't "feel right" in main.js or reducers.js

PR #1018, opened 2016-09-23
refactor(standard): make use of resolve root

joeyhipolito:
It is stated in the README to import modules like so

    import foo from 'components/foo'

But doesn't utilize in the code. I don't know if there is a particular reason why not, so I've created this PR

### DIFF
--- a/src/layouts/CoreLayout/CoreLayout.js
+++ b/src/layouts/CoreLayout/CoreLayout.js
@@ -1,7 +1,7 @@
 import React from 'react'
-import Header from '../../components/Header'
+import Header from 'components/Header'
 import './CoreLayout.scss'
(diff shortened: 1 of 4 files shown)


davezuko:
Thank you for noticing this, and I appreciate the PR, but this is intentional. I know it seems weird: "it's a feature but you're not using it...", but it was done to eliminate the possibility for new users to confuse local modules in the root of the source directory with npm modules, _especially_ where redux is considered.

At first we used relative paths for everything _except_ redux-related code, but that was inconsistent and equally confusing, so we just standardized it everywhere and opted for documenting the _ability_ to use `resolve.root`.

Anyways, hope that offers some sort of insight. Thanks again!

Issue #863, opened 2016-06-15
compile failure doesn't fail tests


gilesbradshaw:
If I write invalid test code it fails to compile - obviously

however when I run npm test all the tests pass (it doesn't run the invalid code)

I can see the parse errors in the console output however presumably if this was happening on a CI server it would consider the tests passed?  So PRs which created invalid test code would still be repoted as passing (unless I have the wrong end of the stick)

Just wondering if you have any thoughts on making npm test report an error on parse errors of test code..

davezuko:
I believe you're right, and it would be wise to make it fail. I'll try to get a PR up sometime today or tomorrow, unless you have any immediate ideas :).

mealeyst:
@davezuko do we have an update to this? I've been trying to get the tests to fail and I have been having a bear of a time trying to get it to do so.

mealeyst:
If it is of any help, I tried to implement the following webpack plugins in the webpack.config.js file that are listed in [this](https://github.com/webpack/karma-webpack/issues/49) github issue.

mealeyst:
While I am not positive, but it seems that the plugins that were suggested for karma-webpack haven't helped me out yet. I was doing some research and I am going to see if I can track it down again but I think I recall hearing that `require.context()` sets modules as optional so if there is an error, the file isn't loaded.

davezuko:
This should be fixed with the above PR. Thanks!

PR #1027, opened 2016-09-28
<fix>(webpack) Throw compliation errors

mealeyst:
This PR corrects #863 where if a syntax error is encountered in tests,
the process throws an error during compilation. Tests are never run and
CI builds should now fail if there are syntax errors. This is done by
adding a small webpack plugin that checks for errors during compilation,
and adds them to an array of errors. This array is then thrown but only
if there are errors during the compilation stage.

Based on the suggestions from:
https://gist.github.com/Stuk/6b574049435df532e905

### DIFF
--- a/build/webpack.config.js
+++ b/build/webpack.config.js
@@ -43,9 +43,33 @@ webpackConfig.output = {
 }
 
 // ------------------------------------
+// Externals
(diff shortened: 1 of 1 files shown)


davezuko:
Looks good, thank you!

PR #1028, opened 2016-09-29
Add COUNTER_DOUBLE_ASYNC action

dciccale:
removed

> NOTE: This is solely for demonstration purposes. In a real application,
>     you'd probably want to dispatch an action of COUNTER_DOUBLE and let the
>     reducer take care of this logic

the change is not that big, and demonstrate what the pattern should be in a real application

### DIFF
--- a/src/routes/Counter/modules/counter.js
+++ b/src/routes/Counter/modules/counter.js
@@ -2,6 +2,7 @@
 // Constants
 // ------------------------------------
 export const COUNTER_INCREMENT = 'COUNTER_INCREMENT'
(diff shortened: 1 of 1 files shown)


davezuko:
Changes look good, but curious why this should be the approach vs. reusing a singular action?

dciccale:
I think it clearly demonstrates the two actions. instead of demonstrating how to reuse code

dciccale:
I also followed what the previous comments said:

> NOTE: This is solely for demonstration purposes. In a real application,		
  you'd probably want to dispatch an action of COUNTER_DOUBLE and let the		
  reducer take care of this logic.

And that's what is actually being done

davezuko:
Works for me!

PR #1032, opened 2016-10-02
<fix>(README.md): correct mispelling of "Spencer Dixon"

server-monitor:


davezuko:
Good catch, thank you!

Issue #160, opened 2015-11-04
Implement application configuration base on build environment


imanhodjaev:
Hi,

I will be very happy to implement this improvement so users can use environment based application (in browser) configuration 

```es6
import config from 'app-config';
```

Possible points for implementation need to modify https://github.com/davezuko/react-redux-starter-kit/blob/master/config/index.js and change this

1. Create a folder with application configuration in `config` directory,
2. Add  `config.set('app_config', 'config/application');`,
3. Add `app_config      : project.bind(null, config.get('app_config'))` to `paths`

Change this
```es6
config.set('utils_aliases', [
  'actions',
  'components',
  'constants',
  'containers',
  'layouts',
  'reducers',
  'routes',
  'services',
  'styles',
  'utils',
  'views'
].reduce((acc, dir) => ((acc[dir] = paths.src(dir)) && acc), {}));
```

to 

```es6
config.set('utils_aliases',
  [
    'actions',
    'components',
    'constants',
    'containers',
    'layouts',
    'reducers',
    'routes',
    'services',
    'styles',
    'utils',
    'views',
    'config'
  ].reduce(
    (acc, x) => {
      if (x === 'config') {
        return ((acc[x] = paths.app_config(env)) && acc);
      }

      return ((acc[x] = paths.src(x)) && acc);
    },  {}
  )
);
```

Or to

```es6
let module_alases = [
    'actions',
    'components',
    'constants',
    'containers',
    'layouts',
    'reducers',
    'routes',
    'services',
    'styles',
    'utils',
    'views'
].reduce((acc, dir) => ((acc[dir] = paths.src(dir)) && acc), {}));

module_alases['app-config'] = paths.app_config(dir);

config.set('utils_aliases', module_alases);
```

What do you think please advise.

davezuko:
Interesting idea @imanhodjaev, but what about just adding a `config` alias and then allowing people to just do:

`import config from 'config'`

Where the file is `~/src/config.js` or `~/src/config/index.js`. I understand where you're coming from with the fact that the app config lives inside of `~/config`, but that config is specifically for the project and build system, and as a result I think it would be confusing to have application code sitting outside of `~/src`.

imanhodjaev:
@davezuko sure we can keep application configuration inside `src` directory.
I am agree with you that keeping configuration outside of `src` will confuse people.
[...]

PR #1053, opened 2016-10-19
Add application configuration based on build environment

thomasthiebaud:
This is a PR related to this [issue](https://github.com/davezuko/react-redux-starter-kit/issues/160) #160 

The main idea is to define an alias into `webpack.config.js` which pointed to the right config file.

It is then as simple as 

```js
import config from `config`
config.my_variable
```

to access the variable


thomasthiebaud:
Does someone checked this PR ?

davezuko:
@thomasthiebaud I'm just getting back into the swing of this project. I've glanced at it, and appreciate the contribution, but it may take me a day or two to come back to considering it more thoroughly. Sorry for the delay!

thomasthiebaud:
I just rebased this PR, so it is up-to-date

kbariotis:
Hello there @thomasthiebaud. Using the alias command means that the imported object will be included as is in the generated assets. Isn't this a problem because even from the configuration that are not being used will be exposed to the consumer.

I am using DefinePlugin to expose environment based configuration to my app, which is being evaluated in compile time and exposes only the required values.

Would you consider a DefinePlugin based solution?

PR #1054, opened 2016-10-19
chore(reduxDevTool): update reduxDevTools to use new api

TheSharpieOne:
`window.devToolsExtension` is deprecated in favor of `window.__REDUX_DEVTOOLS_EXTENSION__` and `window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__`.
See: zalmoxisus/redux-devtools-extension#220

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -27,8 +27,8 @@ let render = () => {
 // Developer Tools Setup
 // ========================================================
 if (__DEV__) {
-  if (window.devToolsExtension) {
-    window.devToolsExtension.open()
+  if (window.__REDUX_DEVTOOLS_EXTENSION__) {
+    window.__REDUX_DEVTOOLS_EXTENSION__.open()
   }
 }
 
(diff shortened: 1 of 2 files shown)


davezuko:
Thank you!

src/main.js
```diff
@@ -27,8 +27,8 @@ let render = () => {
 // Developer Tools Setup
 // ========================================================
 if (__DEV__) {
-  if (window.devToolsExtension) {
-    window.devToolsExtension.open()
+  if (window.__REDUX_DEVTOOLS_EXTENSION__) {
+    window.__REDUX_DEVTOOLS_EXTENSION__.open()
```
davezuko: Ah, very good point @zalmoxisus . I will remove it!

davezuko: Fixed: https://github.com/davezuko/react-redux-starter-kit/commit/2675bb29115b44c3fb077e7359fd88d5e2989d0e

SimonSelg: Yep, if i recall correctly it was meant to open the devtools window in case the "debug in new window" option was enabled. Good catch @zalmoxisus.

PR #1059, opened 2016-10-24
Fixed issue 'Unexpected key counter found in previous state received …

shendepu:
…by the reducer' when added window.__INITIAL_STATE of { counter: 1 }. The counter reducer is async injected, so there no reducer to key 'counter' when combineReducers in createStore intially. Therefore a warning from combineReducers shows in non-production mode, but the initial state of { counter: 1 } is removed in production mode


fdlk:
Nice! We were also running into this issue. Took us a long time to debug.

Issue #1057, opened 2016-10-22
Upgrade to webpack 2


jonricaurte:
Now that webpack 2 is pretty much done and the only thing being worked on is documentation, would be great to update this repo to use it.

kolpav:
Yes I am really looking forward to tree-shaking feature.

cypherq:
@davezuko any plans regarding this one? I can prepare PR, question is if you're interested on migrating right now?

davezuko:
@cypherq I am interested, and would gladly accept a PR. I have not had much time to devote to this project, but it is definitely a worthwhile change.

davezuko:
Since it's the only open PR for it at the moment, moving discussion to https://github.com/davezuko/react-redux-starter-kit/pull/1063. Thanks!

PR #1063, opened 2016-10-25
feat(compile): upgrade to webpack 2

mgmcdermott:
https://gist.github.com/sokra/27b24881210b56bbaff7 may be useful to check out

Having done this a few times, I thought it would be good to upgrade this popular starter kit to webpack 2 (#1057). Several things are going on here to upgrade:
- postcss and sass loader configs have moved into the webpack.LoaderOptionsPlugin (see https://github.com/postcss/postcss-loader/issues/99 for an example)
- module.preLoaders does not exist and has been replaced with the `enforce: 'pre'` option on a regular loader.
- `devtool: null` is no longer valid and has been replaced with `devtool: false`
- karma-webpack-with-fast-sourcemaps is very outdated (~60 commits behind) and does not seem to support webpack 2 so I've removed it in favor of the official karma-webpack
- upgrading of the ExtractTextPlugin is also required for webpack 2. It is only slight changes in config syntax
- `resolve.root` no longer exist and both the diff below and the gist above show the differences

I confirmed that the npm scripts still function as expected.


mgmcdermott:
note that I will be editing the travis.yml to debug the travis build.

mgmcdermott:
it appears that this does not work with cached node_modules. I will dig into what dependency is cached and causing the issues

mgmcdermott:
It seems like the build is failing with an issue related to loader-utils
https://github.com/s-panferov/awesome-typescript-loader/issues/241

Removing the cache should solve the problem.

mgmcdermott:
The following error keeps showing up in travis:
```
Module build failed: Error: parseQuery should get a string as first argument
    at Object.parseQuery (/home/travis/build/davezuko/react-redux-starter-kit/node_modules/loader-utils/index.js:68:9)
    at Object.module.exports (/home/travis/build/davezuko/react-redux-starter-kit/node_modules/babel-loader/index.js:80:35)
```

It seems to be an issue with loader-utils that is solved by re-installing your node_modules which resetting the cache will fix

jonaswindey:
Any reason why you've put `dev_tool` to null?

davezuko:
Thank you for the PR, this is definitely on my radar.

Referencing https://github.com/davezuko/react-redux-starter-kit/issues/1057 for transparency.

mgmcdermott:
@jonaswindey, I changed the `dev_tool` from null to false due to the addition of config validation. Null is invalid. https://github.com/webpack/webpack/blob/996fdc5eebcd5a109b67efb2d53e6639d93fb264/schemas/webpackOptionsSchema.json#L632-L644

jonaswindey:
@mgmcdermott ok, just wondering since the starter kit currently uses compiler_devtool from /config and not null

davezuko:
Thank you for your PR, and apologies for my inattention. I'll be continuing this off of https://github.com/davezuko/react-redux-starter-kit/pull/1091 which builds on your work. Thanks!

Issue #1037, opened 2016-10-05
npm test fails in Ubuntu or Ubuntu in docker container


kyen99:
With a new clone of react-redux-starter-kit (c3abb2), npm test fails when run in a docker container with the official node image (tried 4.6, 6.6 and 6.7). Running npm test on the same source, locally on my mac, works, though (with node 6.7). When run in the docker container it seems to do some bundling with webpack. When run locally on my mac it does not, even with the same NODE_ENV=test in the environment.

Besides the error in the docker container, I can't understand why the npm test runs differently in the container vs outside the container. The only obvious difference I can see is linux vs mac.

Also, running npm test:dev does work in the docker container. It doesn't do any of the webpack stuff, same as it behaves on my mac.

Here is the output when run in the docker container...

```
root@e5c8d3eef1d8:/usr/src/app# npm test
npm info it worked if it ends with ok
npm info using npm@2.15.9
npm info using node@v4.6.0
npm info pretest react-redux-starter-kit@3.0.0-alpha.2
npm info test react-redux-starter-kit@3.0.0-alpha.2

> react-redux-starter-kit@3.0.0-alpha.2 test /usr/src/app
> better-npm-run test

running better-npm-run in /usr/src/app
Executing script: test

to be executed: node ./node_modules/karma/bin/karma start build/karma.conf 
  app:config Creating default configuration. +0ms
  app:config Looking for environment overrides for NODE_ENV "test". +3ms
  app:config No environment overrides found, defaults will be used. +2ms
  app:webpack:config Creating configuration. +675ms
  app:webpack:config Apply ExtractTextPlugin to CSS loaders. +98ms
  app:karma Creating configuration. +2ms

START:
Hash: a912ce28afb7ffdd8ed7
Version: webpack 1.13.2
Time: 43291ms
                                                         Asset       Size  Chunks             Chunk Names
                                         tests/test-bundler.js    2.53 MB       0  [emitted]  tests/test-bundler.js
    tests/test-bundler.js.50313958a3b619afc3f24e733292a1ab.css    8.08 kB       0  [emitted]  tests/test-bundler.js
                                     tests/test-bundler.js.map    3.04 MB       0  [emitted]  tests/test-bundler.js
tests/test-bundler.js.50313958a3b619afc3f24e733292a1ab.css.map    19.7 kB       0  [emitted]  tests/test-bundler.js
                                                   favicon.ico    24.8 kB          [emitted]  
[...]

PR #1064, opened 2016-10-25
Exclude /node_modules/ from webpack in karma config to eliminate erro…

kyen99:
…rs when running tests on linux. For some reason karma with webpack on linux behaves differently (than macOS) when traversing packages during testing. On linux (tested with Ubuntu 14) it generates errors from dependencies of dependencies. Adding an exclude property of the webpack config within the karma config prevents the transversal that causes the errors. After adding this exclude, both linux and macOS behave the same. An issue with some related notes can be found here: https://github.com/davezuko/react-redux-starter-kit/issues/1037


davezuko:
Thank you!

PR #1086, opened 2016-11-19
fix(babel-core): version conflict between "dependencies" and "devDependencies"

baopham:


davezuko:
Hey, I just happened to notice this earlier as well. Thanks for the fix, but it's probably better that we remove the duplicate dependency. This has been done in https://github.com/davezuko/react-redux-starter-kit/commit/a75570e8b667e63ea79df8eca25782dfe3ad67cd. Thank you for your help!

PR #1103, opened 2016-12-10
A little more beautiful

webmalex:
To be able to move components in subordinate Route


davezuko:
Thanks for the PR! Unfortunately, this is done intentionally; [this comment](https://github.com/davezuko/react-redux-starter-kit/pull/1018#issuecomment-249183289) may provide some explanation as to why.

PR #1114, opened 2016-12-19
fix(karma): karma conf filename

dciccale:


davezuko:
Hey, this is actually intentional to make all the config files the same. Good eye, though!

PR #1115, opened 2016-12-20
Changed import of CoreLayout in route index

jakehm:
The point of CoreLayout/index.js is to import it instead of CoreLayout.js right?  It is CoreLayout.js's container.


mikedevita:
👍 its one of the first things i do to this kit every time i clone it lol.

davezuko:
Doh, makes total sense; just an oversight. Thank you!

PR #1136, opened 2017-01-12
Use export from rather than import and export

jasonallen:
This is a pretty minor improvement, but I thought it might help since it's likely many folks will be reusing/copying this pattern. Instead of doing an `import` and `export`, why not shorten it to just `export from`.


jasonallen:
grrr - need to use `export default from ...` rather than named export. I'll resubmit with that later...

PR #1141, opened 2017-01-17
fix(postcss.config.js): add empty postcss.config.js file so tests can…

prestonbernstein:
… be successfully run.


davezuko:
I believe this needs to be fixed in the webpack config itself, but this at least fixes the immediate issue. Thank you!

PR #1144, opened 2017-01-20
Add AggressiveMergingPlugin to Webpack config

JamesWatling:
I ran into an issue with this starter kit once my project became relatively large. The build step would never complete and would run forever, after days of debugging the simplest and most elegant solution was to add the AggressiveMergingPlugin.

It also speeds up my production deploy from 12 minutes to 3 minutes

Thanks


davezuko:
Nice addition, thank you!

Issue #1132, opened 2017-01-09
state.location is null on page load


davecarlson:
When navigating directly to a subroute (first load of an app), state.location is null.

However, on navigating to a url in the app, the location is filled correctly:

![screen shot 2017-01-09 at 3 15 42 pm](https://cloud.githubusercontent.com/assets/299702/21771264/837ff272-d67e-11e6-9ee2-71f5a712d7bf.png)

Is there a way to push this on page load ?

davecarlson:
I've temporarily solved this by updating `store/location.js` to include the following, but hopefully there's a more "correct" way:

BEFORE:
```
const initialState = null
```

AFTER
```
const getParams = query => {
  if (!query) {
    return { }
  }

  return (/^[?#]/.test(query) ? query.slice(1) : query)
    .split('&')
    .reduce((params, param) => {
      let [ key, value ] = param.split('=')
      params[key] = value ? decodeURIComponent(value.replace(/\+/g, ' ')) : ''
      return params
    }, { })
}

// ------------------------------------
// Reducer
// ------------------------------------
const initialState = {
  pathname: window.location.pathname,
  search: window.location.search,
  hash: window.location.hash,
  action: null,
  key: null,
  query: getParams(window.location.search)
}
````

**Comment by code-press** [comment]
I am importing `browserHistory` in location.js and setting `initialState` like so:

```
// ------------------------------------
// Reducer
// ------------------------------------
const initialState = browserHistory.getCurrentLocation()
export default function locationReducer (state = initialState, action) {
  return action.type === LOCATION_CHANGE
    ? action.payload
    : state
}
```

davecarlson:
@code-press  - Yup, that works well.

Thanks!

davecarlson:
Created a PR with the change - https://github.com/davecarlson/react-redux-starter-kit/commit/a67494f0c554e228b0c93c6933b3413fcdd6331d

davezuko:
Fixed with your PR, thank you.

PR #1149, opened 2017-01-23
Initial State - refs #1132

davecarlson:


Piu130:
Looks nice. Just the two tests fails because they expect initial state = null.

davecarlson:
Ah yes I'll fix those up

davecarlson:
Have updated the PR with tests, and confirm travis has now passed.

davezuko:
Thank you, this looks great. Good catch.

PR #1154, opened 2017-01-26
Prefer es6 module loaders over mixing it with commonjs because of fea…

OskarKaminski:
Prefer es6 module loaders over mixing it with common.js because of features like statical code analyse and tree shaking (soon)

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -2,6 +2,7 @@ import React from 'react'
 import ReactDOM from 'react-dom'
 import createStore from './store/createStore'
 import AppContainer from './containers/AppContainer'
+import createRoutes from './routes/index'
 
 // ========================================================
 // Store Instantiation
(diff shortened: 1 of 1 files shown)


src/main.js
```diff
@@ -15,7 +16,7 @@ const store = createStore(initialState)
 const MOUNT_NODE = document.getElementById('root')
 
 let render = () => {
-  const routes = require('./routes/index').default(store)
```
prescottprue: Using require this way waits until it is called to load the dependency. Using imports imports regardless.

davezuko:
As @prescottprue the use of `require` in this codebase is intentional. We always prefer imports when the module can be statically imported, for the benefits you mention, but require is needed (at least for now) for dynamic imports at runtime. Appreciate the PR, though!

Issue #1207, opened 2017-04-02
RedBox error not mapped to original source code


vikas0380:
Error in RedBox/Console showing links to compiled code instead of original code from Webpack  
![screen shot 2017-04-03 at 1 23 46 am](https://cloud.githubusercontent.com/assets/4193675/24590545/7a7ec296-180c-11e7-958a-311a78986c15.png)

davezuko:
Closing in favor of PR discussion.

PR #1208, opened 2017-04-02
#1207 fix(Config,Babel): Redbox error source mapping change

vikas0380:
RedBox showing error links to compile code instead of original code from webpack #1207

### DIFF
--- a/config/project.config.js
+++ b/config/project.config.js
@@ -32,10 +32,10 @@ const config = {
   // ----------------------------------
   compiler_babel : {
     cacheDirectory : true,
-    plugins        : ['transform-runtime'],
+    plugins        : ['transform-react-display-name', 'transform-runtime'],
     presets        : ['es2015', 'react', 'stage-0']
   },
-  compiler_devtool         : 'source-map',
+  compiler_devtool         : 'eval',
   compiler_hash_type       : 'hash',
   compiler_fail_on_warning : false,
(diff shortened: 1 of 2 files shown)


vikas0380:
screen shot 2017-04-03 at 1 51 57 am

prescottprue:
+1 for this

I was able to get the intended result without including `'transform-react-display-name'`, so that may not be needed.

davezuko:
Really appreciate the fix, but if it's solvable without the added dependency I'd prefer that route.

Do normal source maps still work correctly with `eval`? I remember having issues in the past, but perhaps those have been fixed.

vikas0380:
@prescottprue  Can you please provide the details, how you were able to do that. Because without adding a dependency is always a preferred solution for me.

@davezuko Very much agreed with you, without adding a dependency is always preferred by me as well. 
i haven't faced any issue  regarding eval with source maps as such. As far as Redbox is concern they recommend eval over others with webpack. https://github.com/commissure/redbox-react#sourcemaps-with-webpack

davezuko:
So I just took a look at this with the new migration to webpack 2, and it seems like it's just an issue with Redbox. If you look at the error in the console, you see the correct component and file mapping in the stack trace. Going to dig into redbox and see about submitting a PR to them, since `source-map` (and similar) is more useful for debugging than `eval`.

PR #1236, opened 2017-05-08
Adding in useful git hooks

olliecaine27:
1. Run unit tests before Git push.
2. Retrieve new Node modules after Git pull.


olliecaine27:
Thanks for the feedback!

PR #1245, opened 2017-05-17
Fix Node Support

davezuko:


davezuko:
@richyrb00 I just tested this locally with Node 5 (minimum supported version for this project) and the changes worked for all commands. Also tested with Node 4 (technically not supported, but was curious) and that worked as well, so you should be good. Added Node 5 to the CI build so any issues like this get caught in the future.

Sorry about that, and thanks for the quick find.

PR #1247, opened 2017-05-19
fix(package.json): build script for windows env

osmanov:
I use windows machine,when I tried to build my app I had an issue with NODE_ENV.This changes fixed it.


davezuko:
Oh, whoops, my bad. Good catch, thank you.

PR #1258, opened 2017-05-29
Use import() instead of require.ensure for async split

tusbar:
https://webpack.js.org/guides/code-splitting-async/


davezuko:
Thanks for the PR. I've currently been holding off due to https://github.com/webpack/webpack/issues/4636#issuecomment-301920118, but will take another look through related discussions and see if there's been any improvement.

joeyhipolito:
I believe you can consider this now @davezuko :)

Issue #1224, opened 2017-04-19
Unit testing with Jest


kwnccc:
Jest seems a really nice solution as far as testing both components and redux parts (action creators, reducers, etc). Also running tests using jsdom is much faster compared to Karma that needs a browser to run the tests. I would like to make a PR in order to replace Karma + Mocha + Chai + Sinon with Jest + Enzyme. Just wondering, is this is fine with the project objectives?

davezuko:
Been away from this project for a while. I have not used Jest very much, but would be willing to give it a shot. I do prefer JSDOM over PhantomJS for the reasons mentioned, but do appreciate the safety offered by running tests in a legitimate browser.

If you'd like to submit a PR I can promise I'll seriously consider it. If not, I'll try to learn Jest in my spare time and evaluate a migration.

kwnccc:
I'd really like to submit a PR, I believe that there is a great potential in testing with Jest. Will come back to you with the PR related to this opened issue.

PR #1265, opened 2017-06-03
Refactor boilerplate jest

kwnccc:
Having replaced Karma with mocha - chai - sinon testing utilities with Jest, this PR includes:
- Jest configuration
- Refactoring tests to Jest
- Addition of snapshot test examples
- Jest coverage & reports
- Updated npm dependencies
- Updated npm scripts
- Removal of previous configuration using Karma with mocha - chai - sinon


kwnccc:
https://github.com/davezuko/react-redux-starter-kit/issues/1224

kwnccc:
@davezuko It seems like codecov fails when comparing with base branch because coverage report has been changed probably. Do you think there is any way to resolve this?

tests/store/createStore.spec.js
```diff
@@ -3,15 +3,11 @@ import {
 } from 'store/createStore'
 
 describe('(Store) createStore', () => {
-  let store
-
-  before(() => {
-    store = createStore()
-  })
+  let store = createStore()
```
boujeepossum: this should definitely use `beforeEach`. otherwise the same store will be reused in multiple tests (leaking state).

or, maybe even better, this variable should be created in every test separately

build/jest.transform.js
```diff
@@ -0,0 +1,4 @@
+var webpackConfig = require('./webpack.config.js')
+var babelConfig = webpackConfig.module.rules[0].use[0].query
```
boujeepossum: it seems like `babelConfig` should be defined in a separate module and required here and in the webpack config  :)

tests/store/createStore.spec.js
leimonio: You are definitely right on this, updating!

build/jest.transform.js
leimonio: @dmitriiabramov You're right here, not so clean passing babel config directly from webpack that way. When extracting babelConfig to different file, do you think it outweighs `.babelrc` configuration file, or should `.babelrc` be used instead? Though, `.babelrc` configuration cannot be directly used in neither of these config files.

tests/store/createStore.spec.js
```diff
@@ -3,27 +3,23 @@ import {
 } from 'store/createStore'
 
 describe('(Store) createStore', () => {
-  let store
-
-  before(() => {
-    store = createStore()
-  })
-
```
joeyhipolito, line 11: why did you favor declaration of store inside each assertions, rather than having them in `beforeEach`?

boujeepossum: reassigning variables in `beforeEach` is a pretty dangerous pattern. You can forget to reassign it and the value will leak into other tests in this file

feedingaliencat:
Tests are broken because of the following error:
`SyntaxError: Unexpected token import`

@kwnccc please consider the PR, that fixes the issue https://github.com/kwnccc/react-redux-starter-kit/pull/1

Issue #1269, opened 2017-06-08
HMR fails on second update


matgessel:
I was having this problem on a derivative project, however I can't reproduce it in the vanilla starter project.

Webpack 2 projects require that `modules: false`be specified for Babel. It appears that this setting is being ignored because it is a property of the wrong object. If you look at the docs for [babel-preset-env](https://www.npmjs.com/package/babel-preset-env#modules), `modules` is a sibling (not a child) of `targets`. Moving the property up one level fixed my issue.

Keywords: webpack webpack2 modules, hot module reloading hmr second, 2nd

PR #1270, opened 2017-06-08
fix(webpack): bug 1269, HMR fails on second update

matgessel:
See #1269 for issue details


davezuko:
Ah, dang, good catch. Thank you for discovering this.

PR #1274, opened 2017-06-13
Remove duplicated export

ossan-engineer:
I'd like to remove duplicated export to fix ESLint error.


ossan-engineer:
I noticed that I was using airbnb-style lint rules, so close this pull request.

Issue #61, opened 2015-09-07
Poll: Keep or Ditch Support for Non-JS Assets?


davezuko:
General question since I don't know how much people are even using this starter kit with non-JS assets (i.e. it's currently setup to allow Sass imports). Basically, do we want the focus of this starter kit to be strong support for client-side development, or should it be a more robust universal kit?

**Benefit of going JS only**
* The server would be able to directly import all application source code, thereby eliminating the build step that's currently required to perform server-side rendering. See: https://github.com/davezuko/react-redux-starter-kit/issues/57 for more on this.
* Server development would be responsible for proxying the webpack assets, so you'd now have a more cohesive development environment.

**Downside of going JS only**
* Would no longer be able to use folder aliases.
* (Obvious) only JavaScript assets would be able to be imported. You'd have to implement your own build step for compiling styles and whatever else you might need.
* There'd need to be some work to get the server to support hashed asset names.

I probably missed something somewhere, but those are my initial thoughts. Any other perspectives are greatly appreciated.

nhagen:
I think another options (which requires the most effort...) is to provide support out of the box for non-js assets and provide some documentation on how to remove them. I think a common reason for someone to use a starter kit is because they tried to configure everything independently and didn't get it quite right, so they went to the kit that had everything they needed. If you're unfamiliar with webpack, for example, its probably easier to strip stuff out than to add stuff. This is especially true if the ecosystem is new and you don't even know *what* you need.

davezuko:
@nhagen I agree that would be nice, but the question is more about whether or not the starter kit should focus on a well-rounded development environment for client code or a more cohesive client+server kit. A lot of the features this starter kit uses are Webpack-specific (aliases, non-JS imports, etc.), so if we want to support them at all (whether or not we document how to remove them) that means the server code must either be bundled with Webpack, or use a Webpack'd application bundle (which is what it currently does).

From what you say, I'm leaning more towards keeping the server separated from the application source code, that way the starter kit can provide more options to those who, as you say, may not know how to configure it themselves.

mlusetti:
I totally agree on keeping the server separate and I actually like more the support for client-side development.
An universal kit is something everyone will strive to but to achieve it you've to be more opinionated.
I find perfectly suitable to have a webpack'd application bundle.

tsemerad:
[...]

PR #1307, opened 2017-07-11
Adding code coverage using Istanbul

sproogen:
Really useful starter kit, but I had noticed code coverage had been disabled in the latest version and I ran into a number of issues trying to get it included again.

In the end I managed to get it working with `karma-coverage-istanbul-reporter` and `istanbul-instrumenter-loader`.

I'm not sure if this is the perfect solution but thought I would pull request this solution back in.


prescottprue:
When implementing it worked great, but I got the following deprecation message:

deprecation warning from loader utils

To keep this warning from appearing I needed to update `istanbul-instrumenter-loader` to `v3.0.0-rc.1` (can be installed using `npm i --save-dev istanbul-instrumenter-loader@beta`). It seems like everything beyond `v3.0.0-beta.1` should work since it was fixed in [issue #61 of istanbul-instrumenter-loader](https://github.com/webpack-contrib/istanbul-instrumenter-loader/pull/61).

Note that this was just to fix a deprecation warning from `loader-utils`, so if you don't mind getting that warning, then the stable version is just fine.

sproogen:
That's good to know, I couldn't work out how to remove the deprecation messages.

Hopefully they will release `v3.0.0` of the `istanbul-instrumenter-loader` soon then.

Thanks

